PR #3, opened 2020-01-24
downgrade to dune 1 and make jest work (partially)

jchavarri:
- Downgrades to dune to `bin.exe` is generated
- Starts plugging Jest tests


test_bs/esy.json
```diff
@@ -1,13 +1,10 @@
 {
   "name": "test_bs",
-  "version": "0.0.0",
+  "version": "0.0.1",
```
jchavarri: lol

PR #5, opened 2020-01-25
Try to render ast

davesnx:
This parses CSS and returns the AST (no clue what's the schema of it), but we will need to map that to https://github.com/ahrefs/bs-emotion/blob/master/bs-emotion/src/Emotion.re

I removed "mode" from `css_to_emotion.re`, just to start understand a little the file.

### DIFF
--- a/lib/css_to_emotion.re
+++ b/lib/css_to_emotion.re
@@ -6,7 +6,7 @@ open Parsetree;
 open Css_types;
 
 let grammar_error = (loc, message) =>
-  raise([@implicit_arity] Css_lexer.GrammarError(message, loc));
+  raise(Css_lexer.GrammarError((message, loc)));
 
 let split = (c, s) => {
   let rec loop = (s, accu) =>
@@ -114,7 +114,7 @@ let group_params = params => {
 let is_time = component_value =>
   Component_value.(
     switch (component_value) {
-    | [@implicit_arity] Float_dimension(_, _, Time) => true
+    | Float_dimension((_, _, Time)) => true
     | _ => false
     }
   );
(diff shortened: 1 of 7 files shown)


lib/css_to_emotion.re
```diff
@@ -604,7 +507,7 @@ let rec render_component_value =
     }
   | Unicode_range(_) => grammar_error(loc, "Unsupported unicode range")
   | [@implicit_arity] Function(f, params) => render_function(f, params)
-  | [@implicit_arity] Float_dimension(number, "ms", Time) when mode == Bs_css =>
+  | [@implicit_arity] Float_dimension(number, "ms", Time) =>
```
jchavarri: The `implicit_arity` annotations are artifacts from refmt and could be removed. This could be replaced with `Float_dimension((number, "ms", Time))` (note double parents around tuple args).

lib/re_styled_ppx.re
```diff
@@ -6,8 +6,35 @@ open Parsetree;
...
+        let ast =
+          Css_lexer.parse_string(
+            ~container_lnum,
+            ~pos=loc_start,
+            str,
+            Css_parser.declaration_list,
+          );
+
+          Console.log(ast);
+        Ast_helper.Exp.constant(Pconst_integer("42", None))
```
jchavarri: It feels like this is almost at a point where this line could be replaced with `Css_to_emotion.render_stylesheet(ast)` and it would work?

package.json
```diff
@@ -20,23 +20,20 @@
...
-    "@opam/merlin": "*",
-    "ocaml": "~4.6.1000"
-  },
+  "resolution": { "@opam/sedlex": "ocaml-community/sedlex" },
+  "devDependencies": { "@opam/merlin": "*", "ocaml": "~4.6.1000" },
   "scripts": {
     "test": "dune runtest",
     "test:promote": "dune runtest --auto-promote",
-    "test:regen": "sh -c \"refmt ./test/simple_src.re -p ml > ./test/simple.ml\"",
+    "test:regen":
```
jchavarri: This could be removed for now as there is no simple_src anymore.

lib/css_to_emotion.re
```diff
@@ -5,101 +5,9 @@ open Asttypes;
...
-  | Bs_css => false
-  | Bs_typed_css =>
-    switch (declaration) {
-    | "unsafe" => false
-    | _ => true
-    }
-  };
-
-let is_variant = (mode, ident) =>
-  /* bs-css uses polymorphic variants to enumerate values. Some of them
```
jchavarri: `is_variant` or something similar might be needed for bs-emotion as well, as many values are variants, e.g. https://github.com/ahrefs/bs-emotion/blob/master/bs-emotion/src/Emotion.rei#L681

Issue #44, opened 2020-03-11
Support all HTML attributes for all the tags


davesnx:
Once we finish I#26, we will need to add support to pass props to those components

Issue #26, opened 2020-03-04
Support a way to define any HTML tag


davesnx:
First I thought the syntax would work like that:

- `[%styled.div ""];` But there are a few limitations: Can't define dynamically the HTML tag. Add noise to the api, and It's a little bit complex to create those ppxs. Allows that  It would remove the ability to `styled(Component)`.
- In styled-components they use "as" prop in the render. I like that and is reasonable simple to implement in the ppx.
- I don't have many other ideas, so, feel free to discuss here.

davesnx:
We don't have many alternatives on the ppx. I thought about having `[%styled div "opacity: 1"]`, but ppx's only supporting one parameter on the payload.

Issue #12, opened 2020-03-02
Create dynamic components


davesnx:
Right now we only support components defined with "static" CSS. Where "static" means only working with strings or multi-line strings:

```re
module ComponentMultiline = [%styled
  {|
    color: red;
    background-color: white;
    margin: auto 0 10px 1em;
    border-bottom: thin dashed #eee;
    border-right-color: rgb(1, 0, 1);
    width: 70%;
    background: url(http://example.com/test.jpg);
  |}
]
```

Dynamic would be something like, instead of receiving a string, we would need to add a case where the ppx checks for a function and tries to create the make fn with the props that are being passed to that fn.

```re
module StyledComponentWithProps = [%styled
  (~color) => {|
    color: $color;
  |}
];
```

Would turn to be:

```re
module StyledComponentWithProps = {
  let styled = (~color) => Emotion.(css(color $color));

  [@react.component]
  let make = (~color, ~children?) => {
    <div className=styled(~color)>
      {switch (children) {
        | Some(c) => c
        | None => React.null
      }}
    </div>
  }
];
```

As a reference, we would need to do similar logic from react's ppx in order to create the styled and make fns.

Here is the jsofocaml:
https://github.com/jchavarri/jsoo-react/blob/master/ppx/Rroo_jsoo_ppx.re#L213-L658

davesnx:
- [ ] Change to transform styled with functions as well.
- [ ] Turn styled to `let styles = () => Emotion.(css([display(`block)]))`
- [ ] Change parser to accept variables in reason ($things).
- [ ] transform input fn to styles fn
- [ ] optimize, split static properties from dynamic properties and cache the static ones.

PR #46, opened 2020-03-11
Support all HTML tags with all React domProps

davesnx:
> #### Old issue: 
> First I thought the syntax would work like that:
> - `[%styled.div ""];` But there are a few limitations: Can't define dynamically the HTML tag. Add noise to the api, and It's a little bit complex to create those ppxs. Allows that  It would remove the ability to `styled(Component)`.
> - In styled-components they use "as" prop in the render. I like that and is reasonable simple to implement in the ppx.

Hi, 

This PR started as "I want to support all HTML elements", so I added a list and changed the ppx to be `[%styled.div]` and all the variants.

Later I decided to push a little forward to see how hard would be to add the domProps that React uses in their JSX to transform to dom attributes or event handlers internally.

Finally I end up with a rewrite of the module to bypass [@react.component] and just create the makeProps record.

I made it to this PR since it allows to think about the next steps to turn styled into a function and be ready to receive those makeProps. If it's a function we would be ready to get the ast and detect which props are passed to the StyledComponent and create the styled with those props (and add them to makeProps). (Exactly this issue https://github.com/davesnx/styled-ppx/issues/12)

So, this PR it would solve https://github.com/davesnx/styled-ppx/issues/44 and https://github.com/davesnx/styled-ppx/issues/26

In order to merge this PR we need to:
- [x] Create a binding to `React.createElement` https://es.reactjs.org/docs/react-api.html#createelement
- [x] Add a type to the module to hide the `setClassName` and the `React.createElement` binding. https://2ality.com/2017/12/modules-reasonml.html
- [x] Add mutable into className
- [x] Update README.md and ROADMAP.md
- [ ] Fix `styled.tag` ppx detection. Probably worth it adding another solution? https://github.com/paurkedal/ppx_regexp https://github.com/Drup/tyre
- [ ] Try to improve test native.
- [x] Create some snapshot tests in BuckleScript.


src/styled_ppx.re
```diff
@@ -137,12 +259,12 @@ let moduleMapper = (_, _) => {
     switch (expr) {
     | {
         pmod_desc:
-          /* that are defined with the ppx [%styled] */
+          /* that are defined with a ppx like [%txt] */
           Pmod_extension((
-            {txt: "styled", _},
+            {txt, _},
```
davesnx: Hey @jchavarri, do you know if this is performant? I'm mapping thought all the ppx that have a text and later on, doing the split.

So I'm not sure if it's even a good idea. WDYT?

davesnx: I think it blocks other ppx to run actually 😢

davesnx:
Probably worth it adding https://github.com/Drup/tyre to catch up `styled.tag` better.

src/styled_ppx.re
jchavarri: It shouldn't block other ppxs. I think the issue might be that the current code is "swallowing" the attributes of other ppxs.

If the ast returned respects other existing attributes and returns them, other ppxs should work ok 🤔

Issue #49, opened 2020-03-15
Negative number not supported


davesnx:
![Screenshot 2020-03-15 at 19 38 53](https://user-images.githubusercontent.com/3763599/76708109-ae448180-66f4-11ea-8cc0-5b99c55a020e.png)

PR #54, opened 2020-03-20
#12 Dynamic components

davesnx:
https://github.com/davesnx/styled-ppx/issues/12

**In order to support components with styles based on some static variable**
```reason
let b = "flex";
module Component = [%styled.section {j| display: $b |j}];
```
- [x] Change parser to accept interpolate strings in reason like `$things`.
- [x] Change parser to accept interpolate strings in reason like `$(things)`.
- [x] Transform them to valid Emotion calls.

**Components with styles based on props that defines the style def**
```reason
module Component = [%styled.section (~mode) => {j| display: $mode |j}];
module Component = [%styled.section (~num) => {j| margin: $(num)px |j}];
```
- [x] Change to transform styled with functions as well.
- [x] Turn styled to `let styles = () => Emotion.(css([display('block)]))`
- [x] Pass arguments to a the styled() fn.
- [x] simple.ml: `Error: broken invariant in parsetree: Function application with no argument`
- [ ] ~Fix **`Warning 44: this open statement shadows the value identifier color (which is later used)`**. Two options: 1. Remove the open and prefix all func with Emotion. 2. Prefix all custom props with "styledPpx{{Variable}}", this can cause error messages to show up the name.~
- [x] Add customProps to makeProps type def.
- [x] How can I know the type of the prop?. (We will set all to string, since Emotion.p treats second arg as string).
- [x] Revert makeProps
- [x] Reflect documentation
- [x] Update ROADMAP
- [x] Stress test the feature and document findings/bugs.

PD: This PR accidentally solves https://github.com/davesnx/styled-ppx/issues/49


davesnx:
simple.ml: `Error: broken invariant in parsetree: Function application with no argument`

PR #81, opened 2020-05-06
empty %styled module exprs support

an146:


demo/src/index.re
```diff
@@ -28,6 +28,9 @@ module Link = [%styled.div {|
   margin-top: 16px;
 |}];
 
+module Empty = [%styled.span]
+module Empty1 = [%styled ""]
```
davesnx, line 32: Maybe adding those into "simple.ml" where you do the snapshotting as well would be a good idea.

PR #85, opened 2020-05-13
Feature/nested selectors

an146:
seems to work, but code may be not ideal

### DIFF
--- a/src/css_to_emotion.re
+++ b/src/css_to_emotion.re
@@ -605,7 +605,7 @@ let rec render_value = ((cv, loc): with_loc(t)): expression => {
     };
   | String(s) => string_to_const(~loc, s)
   | Selector(s) =>
-    Exp.ident(~loc, {txt: Lident(s), loc});
+    Exp.ident(~loc, {txt: Emotion.lident(s), loc});
   | Uri(s) =>
     let ident = Exp.ident(~loc, {txt: Emotion.lident("url"), loc});
     let arg = string_to_const(~loc, s);
--- a/src/css_to_emotion.re
+++ b/src/css_to_emotion.re
@@ -189,7 +189,7 @@ let string_to_const = (~loc, s) =>
   Exp.constant(~loc, Const.string(~quotation_delimiter="js", s));
 
 let render_html_color = (~loc, v) =>
-  Exp.ident(~loc, {txt: Ldot(Ldot(Lident("Css"), "Color"), v), loc: loc});
+  Exp.ident(~loc, {txt: Ldot(Ldot(Emotion.lident("Css"), "Color"), v), loc: loc});
 
 let list_to_expr = (end_loc, xs) =>
   List.fold_left(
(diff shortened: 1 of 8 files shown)


src/css_to_emotion.re
```diff
@@ -1111,22 +1113,17 @@ and render_declarations =
       | Declaration_list.Declaration(decl) =>
         render_declaration(decl, decl.loc, Option.get(variables, []))
       | Declaration_list.At_rule(ar) => render_at_rule(ar)
+      | Declaration_list.Style_rule(ar) =>
+        let (loc: Location.t) = ar.loc;
+        render_style_rule(Exp.ident(~loc, {txt: Lident("select"), loc}), ar)
```
davesnx, line 1118: Maybe you want to use `Emotion.lident` right?

src/css_to_emotion.re
```diff
@@ -1146,6 +1144,7 @@ let render_style_rule = (ident, sr: Style_rule.t): expression => {
       "",
       List.rev(prelude),
     );
+  Printf.printf("%s", selector);
```
davesnx, line 1147: We should remove this

src/css_to_emotion.re
```diff
@@ -1156,6 +1155,26 @@ let render_style_rule = (ident, sr: Style_rule.t): expression => {
...
 
+
+let render_emotion_css = ((list, loc): Declaration_list.t, variables): expression => {
+  let declarationListValues = render_declaration_list((list, loc), variables);
+  let ident = Exp.ident(~loc, {txt: Emotion.lident("css"), loc});
+
+  Exp.apply(~loc, ident, [(Nolabel, declarationListValues)]);
+};
+
+let openEmotionWrapper = (~loc, expr) => {
```
davesnx, line 1166: I removed openEmotionWrapper in previous PRs in favour of Emotion.lident, son instead of open inline we can use namespace: for example: `Emotion.css`

src/css_to_emotion.re
```diff
@@ -1156,6 +1155,26 @@ let render_style_rule = (ident, sr: Style_rule.t): expression => {
...
+  Exp.open_(~loc, Fresh, {txt: Lident("Emotion"), loc}, expr);
+};
+
+let render_emotion_css = ((list, loc): Declaration_list.t, variables): expression => {
```
davesnx, line 1170: This function seems duplicated, maybe the upper one it's the correct

src/css_to_emotion.re
```diff
@@ -604,6 +604,8 @@ let rec render_value = ((cv, loc): with_loc(t)): expression => {
       Exp.ident(~loc, {txt: Emotion.lident(name), loc});
     };
   | String(s) => string_to_const(~loc, s)
+  | Selector(s) =>
+    Exp.ident(~loc, {txt: Lident(s), loc});
```
davesnx, line 608: I'm sure this Lident should be Emotion.lident as well:

```suggestion
    Exp.ident(~loc, {txt: Emotion.lident(s), loc});
```

PR #92, opened 2020-05-28
Feature/pseudoclasses

davesnx:


src/css_to_emotion.re
```diff
@@ -1126,33 +1125,98 @@ and render_declaration_list = ((list, loc): Declaration_list.t, variables): expr
...
+    | "placeholder" => "placeholder"
+    | _ => grammar_error(loc, "Unexpected pseudo-class")
+    };
+    let ident = Exp.ident(~loc, {txt: Emotion.lident(f), loc});
+    Exp.apply(
+      ~loc=sr.Style_rule.loc,
+      ident,
+      [(Nolabel, dl_expr)],
+    );
+  | [(Selector("&"),_),(Delim(":"),_),(Ident(pc),loc)] =>
```
davesnx, line 1147: What do you think about adding a comment here explaining the difference between this case and the above one?

src/css_to_emotion.re
```diff
@@ -1126,33 +1125,98 @@ and render_declaration_list = ((list, loc): Declaration_list.t, variables): expr
...
+          List.rev(args)
+        )
+      | _ => grammar_error(value_loc, "Unexpected selector")
+      }
+    };
+    let selector =
+      List.fold_left(
+        render_prelude_value,
+        "",
+        List.rev(prelude),
```
davesnx, line 1210: Instead of reverting the list, maybe fold_right?

src/css_to_emotion.re
```diff
@@ -1126,33 +1125,98 @@ and render_declaration_list = ((list, loc): Declaration_list.t, variables): expr
...
+    | "empty" => "noContent"
+    | "default" => "default"
+    | "any-link" => "anyLink"
+    | "only-child" => "onlyChild"
+    | "only-of-type" => "onlyOfType"
+    | "optional" => "optional"
+    | "invalid" => "invalid"
+    | "out-of-range" => "outOfRange"
+    | "target" => "target"
+    | _ => grammar_error(loc, "Unexpected pseudo-class")
```
davesnx, line 1171: nth-child doesn't appear here, but appears on the test. am I missing something here?

PR #94, opened 2020-05-30
Road to `bs-css`

### LINKED ISSUE #77: Migrate from bs-emotion to bs-css-emotion
bs-css-emotion seems more lightweight than bs-emotion and relies on re-css.
bs-css comes with a very handy css_core and with tiny bindings to emotion.

The idea would be to stop relying on having the peerDependency of bs-emotion and bundle bs-css-emotion and bs-css inside styled-ppx.

EduardoRFS:
* improved tests
* migrated to bs-css

* closes #77


davesnx:
Hey @EduardoRFS, If we merge this we should publish a breaking change and update the readme as well.

We would need to take into account this PR as well: https://github.com/davesnx/styled-ppx/pull/92

src/css_to_emotion.re
```diff
@@ -1175,98 +1162,81 @@ and render_style_rule = (ident, sr: Style_rule.t): expression => {
...
     /* nth-child & friends */
-    let f = switch (pc) {
-    | "nth-child" => "nthChild"
-    | "nth-last-child" => "nthLastChild"
-    | "nth-of-type" => "nthOfType"
-    | "nth-last-of-type" => "nthLastOfType"
-    | _ => grammar_error(loc, "Unexpected pseudo-class")
-    };
-    let ident = Exp.ident(~loc, {txt: Emotion.lident(f), loc});
+    // TODO: parses and use the correct functions instead of just strings selector
```
davesnx, line 1236: 👍

PR #95, opened 2020-06-18
Css wide keywords

EduardoRFS:
## What this does?

It updates our dependencies and use `reason-css-parser` to support css wide keywords for all properties.

## Why?

The way that this was implemented isn't actually a requirement as it could easily be done by just using a pattern matching on string, but it integrates the tooling to avoid big changes in future PRs and makes our first integration with `reason-css-parser`.

## ** WARNING **

If you use an invalid property name, it will still accept any of the keywords, so `potato: unset` will work, but that should be a minor as it's better to support all properties even if partially, while we work on proper supporting them. Also that is the same for `potato: $var`


EduardoRFS:
Closing as it was already merger at #95

PR #97, opened 2020-07-04
Implements css-flexbox-1

EduardoRFS:
## What this PR does?

As we now have a new parser for properties, this PR implements a way to integrate properties using the new `reason-css-parser`, this should make a lot easier to integrate new properties.

Also the module for `css-flexbox-1` was implemented as an example

## Problems

It requires a string of input to parse the property value, so a workaround was implemented to recover the input from the original buffer, to make that work access to a copy of the original buffer was needed and as this is temporary a global variable was introduces holding a copy of the current buffer.

- Error messages are still terrible
- Variables are still not supported

I'm already working on both of them, and will open a PR in the following days


parser/lex_buffer.re
```diff
@@ -121,28 +125,8 @@ let utf8 = (~skip=0, ~drop=0, lexbuf) => {
 
 let container_lnum_ref = ref(0);
 
-let fix_loc = loc => {
-  let fix_pos = pos => {
-    /* It looks like lex_buffer.ml returns a position with 2 extra
-     * chars for parsed lines after the first one. Bug? */
```
EduardoRFS, line 127: this workaround wasn't needed(anymore?) and was introducing a bug when combining with the new workaround for properties. As we now have a better test suite this should be safe to change.

src/css_to_emotion.re
```diff
@@ -29,14 +29,6 @@ open Longident;
 open Css_types;
 open Component_value;
 
-module Option = {
-  let get = (opt, def) =>
```
EduardoRFS, line 33: 4.8 introduced the same function

https://caml.inria.fr/pub/docs/manual-ocaml/libref/Option.html

src/css_to_emotion.re
```diff
@@ -1103,61 +1095,85 @@ and render_declaration =
...
-      render_border_outline(newValueList, loc)
-    | _ => render_standard_declaration(fnName, valueList)
-    }
+  let value_source = {
+    let Warnings.{loc_start, loc_end, _} = loc;
+    let Lex_buffer.{buf, pos, _} = Lex_buffer.last_buffer^;
+    let pos_offset = pos.Lexing.pos_cnum;
+    let loc_start = loc_start.Lexing.pos_cnum - pos_offset;
+    let loc_end = loc_end.Lexing.pos_cnum - pos_offset;
+    Sedlexing.Latin1.sub_lexeme(buf, loc_start - 1, loc_end - loc_start);
```
EduardoRFS, line 1144: This is a workaround, the idea is to recover the original input

PR #104, opened 2020-07-15
css-images-4 and css-backgrounds-3

EduardoRFS:
## Not properties change

Switched to ppxlib, it has a lot of helpers and ocaml-migrate-parsetree driver will be removed

## Why so many unsupported?

Most of them aren't actually supported at all in `bs-css`, some of the `background-*` I will support in a future PR, as it is quite a lot of work to implement them, but with #103 it should be working

### DIFF
--- a/src/declarations_to_emotion.re
+++ b/src/declarations_to_emotion.re
@@ -1,12 +1,16 @@
 open Migrate_parsetree;
 open Ast_410;
 open Ast_helper;
+open Asttypes;
 open Longident;
 open Reason_css_parser;
 open Reason_css_lexer;
 open Parser;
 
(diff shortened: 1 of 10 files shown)


src/declarations_to_emotion.re
```diff
@@ -568,6 +580,250 @@ let opacity =
...
+    property_object_position,
+    render_position,
+    [%expr Css.objectPosition],
+  );
+let image_resolution = unsupported(property_image_resolution);
+let image_orientation = unsupported(property_image_orientation);
+let image_rendering = unsupported(property_image_rendering);
+
+// css-backgrounds-3
+let render_shadow = shadow => {
```
davesnx, line 658: That's very neat man.

Issue #14, opened 2020-03-02
VSCode support extension


davesnx:
- [x] Syntax highlight

As seen on the JavaScript world, there's a few extensions that highlight and autocomplete css: 

https://marketplace.visualstudio.com/items?itemName=mf.vscode-styled-components

https://scotch.io/tutorials/create-your-first-visual-studio-code-extension

https://www.youtube.com/watch?v=srwsnNhiqv8

davesnx:
Published under: https://marketplace.visualstudio.com/items?itemName=davesnx.vscode-styled-ppx

PR #108, opened 2020-07-15
Add css highlighting for vscode extension

mnxn:
Partially completes https://github.com/davesnx/styled-ppx/issues/14

Here's an example with default and custom syntax highlighting:
| Default Theme |  Custom Theme |
| - | - | 
| example | example-with-vscode-theme |

Seems to work fine with OCaml syntax too:
| Default Theme |  Custom Theme |
| - | - | 
| example-ml | example-ml-with-vscode-theme |

The only thing I cant find out how to do is to make the source language's keywords highlighted in the extension node, for example, the `fun` in the OCaml syntax isn't highlighted. Using `$self` or `$base` doesn't seem to work in an embedded syntax.


davesnx:
Oh wow, @mnxn Thanks.

I spend too much time with those syntaxes configs without understanding what I was doing. Really appreciate this, you even did all the cases.

Issue #109, opened 2020-07-25
Check if expression text is different from "styled"


vmarcosp:
Styled-ppx is not checking if expression txt is [%styled], e.g:

![Captura de Tela 2020-07-25 às 18 02 12](https://user-images.githubusercontent.com/20327229/88466970-365b4480-cea8-11ea-97ee-62acfd311451.png)

I tried using something like `[%graphql ...]` and it compiles. Taking a look at the code, i saw that there are no validations for that and the function that extracts the tag name from the text, returns "div" to anything that is not "styled":

```reason
let getTag = str => {
  switch (String.split_on_char('.', str)) {
  | ["styled"] => "div"
  | ["styled", tag] => tag
  | _ => "div"
  };
};
```
https://github.com/davesnx/styled-ppx/blob/master/src/styled_ppx.re#L50

davesnx:
Hey @vmarcosp,

It looks like a silly bug regarding the ppx detection, shouldn't be too hard to fix.

Thanks for reporting.

PR #110, opened 2020-07-27
check if text starts with "styled"

EduardoRFS:
This PR adds some guards to prevent us from transforming code not starting with `styled`, as it would conflict with things like `graphql-ppx`

As described by #109

### DIFF
--- a/test/native/lib/TestModule.re
+++ b/test/native/lib/TestModule.re
@@ -1,0 +1,29 @@
+open Migrate_parsetree;
+open Ast_410;
+open Setup;
+
+open Ppxlib.Ast_builder.Make({
+       let loc = Location.none;
+     });
+
+let compare = (result, expected, {expect, _}) => {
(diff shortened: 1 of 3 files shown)


test/native/lib/TestModule.re
```diff
@@ -0,0 +1,29 @@
...
+       let loc = Location.none;
+     });
+
+let compare = (result, expected, {expect, _}) => {
+  let result = Pprintast.string_of_structure([result]);
+  let expected = Pprintast.string_of_structure([expected]);
+  expect.string(result).toEqual(expected);
+};
+
+describe("transform module ppx", ({test, _}) => {
```
EduardoRFS, line 15: This message is too generic, but I couldn't think about anything else, at least now we have a place to test more generated code, maybe even testing the complete module

davesnx: It's good enough

PR #111, opened 2020-07-28
Typechecking for tests

EduardoRFS:
**YEY +3536**

## What this PR does?

The intent is to mostly to replace(or automate) the bucklescript test suite in the future, but also to make it easier to be sure things are working, as there is no need to duplicate tests between the native and bucklescript.

And in the near future(tomorrow?) to integrate the current snapshot tests on the tests at `test/native/lib`

## Problems

I couldn't find a better way to run the typechecker on dune without pointing straight to the `.cmi` file.

## NOTE

I would really recommend you to review by commit

### DIFF
--- a/test/native/bs-css/Js.rei
+++ b/test/native/bs-css/Js.rei
@@ -1,0 +1,2 @@
+module Json: {type t;};
+module Dict: {type t('a);};
--- a/test/native/bs-css/dune
+++ b/test/native/bs-css/dune
@@ -1,0 +1,3 @@
+(library
+ (name StyledPpxTestNativeBSCSS)
+ (modules_without_implementation js css_AtomicTypes css_Core))
--- a/test/native/lib/TestEmitter.re
+++ b/test/native/lib/TestEmitter.re
@@ -19,6 +19,16 @@ let extract_tests = array_expr => {
     ),
   );
 };
+let write_tests_to_file = (tests, file) => {
+  let code =
+    tests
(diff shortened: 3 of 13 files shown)


test/native/bs-css/Js.rei
```diff
@@ -0,0 +1,2 @@
+module Json: {type t;};
+module Dict: {type t('a);};
```
EduardoRFS, line 2: Just type mocks, because bs-css uses it and it's only available at BS

test/native/bs-css/dune
```diff
@@ -0,0 +1,3 @@
+(library
+ (name StyledPpxTestNativeBSCSS)
+ (modules_without_implementation js css_AtomicTypes css_Core))
```
EduardoRFS, line 3: This line is the trick, with it we can typecheck but not build, which is exactly what we want

test/native/lib/TestEmitter.re
```diff
@@ -2,6 +2,34 @@ open Migrate_parsetree;
 open Ast_410;
 open Setup;
 
+let extract_tests = array_expr => {
+  open Ppxlib.Ast_pattern;
+  let fail = () =>
+    failwith("can only extract from array((result, expected))");
+  let payload = pexp_array(many(pexp_tuple(many(__))));
```
EduardoRFS, line 9: This matching from ppxlib is really amazing a lot more easy than matching by hand on the AST, really recommend anyone interested to give it a look at the `ast_pattern.mli`

PR #115, opened 2020-07-31
add nightly release on every commit

EduardoRFS:
## What this PR does?
After every commit to master we will get a nightly release on NPM

## Implementation

This isn't the best implementation, but should be a good start, will test it, but feel free to review

## Possible problems

Maybe another action creating commits? Not sure on that

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,6 +1,8 @@
 name: Release
 
 on:
+  label:
+    types: [created]
   push:
     branches:
       - master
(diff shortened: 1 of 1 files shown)


davesnx:
Hey man, what do you think to have a release each PR instead of each commit to master?

EduardoRFS:
@davesnx The problem of each PR is that bad things can be published, and there is no way to remove things from npm. Also then you cannot track nightly, because maybe nightly isn't working

We could do it tho if we had a private repository, but publishing to npm seems like a problem on the security side of things, and I really don't want github bot saying to everyone update things because something happened at styled-ppx

PR #118, opened 2020-08-02
Add everything

EduardoRFS:
As the parser now uses the data from `mdn-data` and having things unsupported by default is better than not having them, this PR adds everything supported by the parser as unsupported.

Everything = https://github.com/EduardoRFS/reason-css-parser/blob/master/lib/Parser.re#L1681

## Regressions

There is some small regressions, mostly from fully integrated generating `bs-css` code to `unsupported + variables`

## Dead code

Also removed the old render of properties as it's not used anymore


.vscode/settings.json
```diff
@@ -3,7 +3,7 @@
     "editor.formatOnSave": true
   },
   "ocaml.sandbox": {
-    "root": "/Users/davesnx/dev/github/styled-ppx",
+    "root": "/home/eduardo/reason/styled-ppx",
```
davesnx, line 6: Lets fight for having different local dev paths

PR #119, opened 2020-08-03
add support to basic media queries

EduardoRFS:
## Problems

The parser now has a bunch of ambiguities, so there is possible to have some weird edge cases.

`[%styled.global]` still doesn't support media queries

### DIFF
--- a/parser/css_parser.mly
+++ b/parser/css_parser.mly
@@ -78,6 +78,13 @@ at_rule:
         loc = Lex_buffer.make_loc $startpos $endpos;
       }
     }
+  | name = NESTED_AT_RULE; xs = prelude_with_loc; LEFT_BRACE; ds = declarations_with_loc; RIGHT_BRACE {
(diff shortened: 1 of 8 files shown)


parser/css_parser.mly
```diff
@@ -78,6 +78,13 @@ at_rule:
         loc = Lex_buffer.make_loc $startpos $endpos;
       }
     }
+  | name = NESTED_AT_RULE; xs = prelude_with_loc; LEFT_BRACE; ds = declarations_with_loc; RIGHT_BRACE {
+      { At_rule.name = (name, Lex_buffer.make_loc $startpos(name) $endpos(name));
+        prelude = xs;
+        block = Brace_block.Declaration_list ds;
+        loc = Lex_buffer.make_loc $startpos $endpos;
+      }
+    }
```
EduardoRFS, line 87: the ambiguities happens because this lexer has some weird hacks, and this is one of them

Issue #61, opened 2020-03-29
Components with styles based on props that are defined statically


davesnx:
```reason
module Component = [%styled
  (~size) => {
    if (size > 0) {
      "display: flex"
    }
  }
];
```
- [ ] Change the PPX detection, recursively look for any sort of function body that returns a string, turn that as the body of the styled fn.
- [ ] Define an API where you can pass more than one fn?
```reason
module Component = [%styled
  (~size) => {
    if (size > 0) {
      "display: flex"
    }
  }

  (~kind) => switch (kind) {
    | Solid => "padding: 10px"
	| _ => "padding: 2px 4px
  }
];

PR #124, opened 2020-08-25
Use ppxlib for pattern matching

EduardoRFS:
This is an improvement on code quality, the main advantage is only that we don't need to pattern match by hand and also removed some duplicated work.

It's needed to implement #61

### DIFF
--- a/src/styled_ppx.re
+++ b/src/styled_ppx.re
@@ -491,6 +491,26 @@ let createMakeProps = (~loc, extraProps) => {
   );
 };
 
+let match_exp_string_payload = expr => {
+  open Ppxlib.Ast_pattern;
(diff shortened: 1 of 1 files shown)


src/styled_ppx.re
```diff
@@ -491,125 +491,118 @@ let createMakeProps = (~loc, extraProps) => {
...
+    expr,
+    (key, payload, delim) => Some((key, payload, delim)),
+  );
+};
+
+let match_mod_payload = mod_expr => {
+  open Ppxlib.Ast_pattern;
+
+  let pattern =
+    pmod_extension(
```
davesnx, line 518: I didn't that was possible! Well done!

PR #125, opened 2020-08-25
Fix complex type on components

EduardoRFS:
This is a bug, declaring anything other than a simple identifier wasn't working previously, with this behavior we can define any inline type on the component declaration

```reason
  (~color: Css.Types.Color.t) => 
    "color: $(color)"
];
```

### DIFF
--- a/test/bucklescript/src/__snapshots__/index_test.bs.js.snap
+++ b/test/bucklescript/src/__snapshots__/index_test.bs.js.snap
@@ -32,3 +32,12 @@ exports[`ComponentLink should render an <a /> tag 1`] = `
   />
 </div>
 `;
+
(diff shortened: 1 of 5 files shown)


test/bucklescript/src/__snapshots__/index_test.bs.js.snap
```diff
@@ -32,3 +32,12 @@ exports[`ComponentLink should render an <a /> tag 1`] = `
   />
 </div>
 `;
+
+exports[`ComponentWithParameter renders 1`] = `
+<div>
+  <div
+    class="css-1roeevf"
+    color="5194459,FF0000"
```
EduardoRFS, line 40: As @davesnx mentioned that's not desirable, but will merge it because that's non blocking

PR #133, opened 2020-09-10
Hide the reason-css-parser error messages

### LINKED ISSUE #120: Basic error messages
With location raise a compile error on "Invalid property".

EduardoRFS:
As discussed before currently, the reason-css-parser error messages aren't great and they contain too much noise, so until we improve it or find a better solution, is better to have a generic error message instead of showing noise.

Closes #120

## Duplicated?

Same code as #128, but in a local branch to publish a nightly


davesnx (approved):
👍

PR #135, opened 2020-09-25
add type inference for styled components

EduardoRFS:
## Why?

Currently by omitting the type parameter it fallbacks to `string` which forces us to actually type the structure, that wasn't a problem before but now that most type variables are typed that isn't simple

## Example

```reason
// from
  (~color: Css.Types.Color.t, ~theme: [`Light | `Dark]) => {
    "background: blue";
    switch (theme) {
    | `Light => "background-color: #F0F0F0"
    | `Dark => "background-color: #202020"
    };
    "color: $(color)";
  }
];
// to
  (~color, ~theme) => {
    "background: blue";
    switch (theme) {
    | `Light => "background-color: #F0F0F0"
    | `Dark => "background-color: #202020"
    };
    "color: $(color)";
  }
];
```

### DIFF
--- a/test/native/snapshot/pp.expected
+++ b/test/native/snapshot/pp.expected
@@ -2703,6 +2703,898 @@ Css.style([Css.unsafe("display", "block")]);
 
 module Component = {
   [@bs.deriving abstract]
+  type makeProps('var) = {
+    [@bs.optional]
+    ref: ReactDOMRe.domRef,
+    [@bs.optional]
+    children: React.element,
+    [@bs.optional]
+    key: string,
+    [@bs.optional]
+    defaultChecked: bool,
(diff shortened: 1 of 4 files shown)


test/native/snapshot/pp.expected
```diff
@@ -2701,6 +2701,898 @@ module Component = {
 
 Css.style([Css.unsafe("display", "block")]);
 
+module Component = {
+  [@bs.deriving abstract]
+  type makeProps('var) = {
```
EduardoRFS, line 2706: makeProps now will have a type parameter for each component parameter

BTW, snapshots are still useful for seeing the difference in the generated code

davesnx (approved):
That's very hot

PR #136, opened 2020-09-25
initial keyframe support

### LINKED ISSUE #80: Support keyframe and animation
```
let fadeIn = [%styled.keyframe {|
  0% { opacity: 0 }
  100% { opacity: 1 }
|}];

module Modal = [%styled.div {|
  animation-name: $fadeIn;
|}];

module Modal = [%styled.div {|
  animation: $fadeIn ease-in 200ms;
|}];
```

EduardoRFS:
## Missing

Currently the animations properties doesn't support variables in the following form `animation: $(fadeIn) ease-in 200ms` so you're going to make the complete way.

```css
animation-name: $(fadeIn);
animation-timing-function: ease-in;
animation-duration: 200ms;
```

## Issues
- Closes #80

### DIFF
--- a/test/bucklescript/src/index_test.re
+++ b/test/bucklescript/src/index_test.re
@@ -2,6 +2,11 @@ open Jest;
 open Expect;
 open ReactTestingLibrary;
 
+let fadeIn = [%styled.keyframe {|
+  0% { opacity: 0 }
+  100% { opacity: 1 }
+|}];
+
 module Component = [%styled.div {|
   display: flex;
   justify-content: center;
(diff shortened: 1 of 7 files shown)


test/bucklescript/src/index_test.re
```diff
@@ -13,6 +18,9 @@ module Component = [%styled.div {|
 
   font-size: 30px;
 
+  animation-name: $(fadeIn);
+  /* animation: $(fadeIn) ease-in 200ms; */
```
davesnx, line 22: Can we add an error message if they are using the short-hand version?

davesnx (approved):

test/bucklescript/src/index_test.re
EduardoRFS: I think is easier to actually support the full version, it's probably the next development phase, developing props definition supporting variables

PR #173, opened 2021-03-24
Introduce unsafe

rusty-key:
This is quick-and-dirty implementation of `unsafe`. 
@davesnx, I would appreciate if you could look at it and verify that I am digging in the right direction.


EduardoRFS:
@rusty-key not sure if you want the syntax, but there is already ways to do unsafe, but they're uglier than this

rusty-key:
Can you share existing options? Uglier is not a problem

davesnx:
I know there's a way to flag some properties to "unsupported" which makes the property to render Css.unsafe. But I'm not able to have it working :(

EduardoRFS:
Not sure this works with components without parameters, but if it doesn't it should. Because of the change that allows `;` to be used as `@`, or in ReScript just the space itself, this means that you can mix strings of css with direct calls to Css module

```reason
  (~color: Css.Types.Color.t) => {
    "background: blue";
    [Css.unsafe("color", "trust-me")]
  }
];
```

rusty-key:
Not sure I understand. Does it work only in rescript syntax? I am getting syntax error: 
```
[1/7] Building src/components/Greet.reast
FAILED: src/components/Greet.reast 
/Users/rustykey/code/exp/try-styled-ppx/node_modules/bs-platform/darwin/bsc.exe  -warn-error +5 -color always -bs-jsx 3 -ppx /Users/rustykey/code/exp/try-styled-ppx/node_modules/@davesnx/styled-ppx/ppx -ppx /Users/rustykey/code/exp/try-styled-ppx/node_modules/react-rules-of-hooks-ppx/Bin.exe -bs-super-errors -bs-no-version-header -o src/components/Greet.reast -bs-syntax-only -bs-binary-ast /Users/rustykey/code/exp/try-styled-ppx/src/components/Greet.re
File "/Users/rustykey/code/exp/try-styled-ppx/src/components/Greet.re", line 5, characters 43-44:
5 |     "background: blue" @ [Css.unsafe("color", "trust-me")];
                                               ^
Error: Parse error while reading token 'EOF'
```

rusty-key:
On a syntax itself: having this syntax (in any form) for a regular style block would be really helpful at this stage of the tool. 

The main reason we cannot start experimenting with it in a real project, is amount of friction required for working around some know issues (like problems with `grid-column-template` values). 

I would just make it much less convenient and much more greppable though, `~` is just a quick solution. Maybe something like `UNSAFE_color: trust-me"`

EduardoRFS:
@rusty-key nope that is supported by reason, we even have tests for it

https://github.com/davesnx/styled-ppx/blob/9451b3725755406554bb2c353c7cc17c7d8bdae0/packages/ppx/test/bucklescript/src/dynamic_test.re#L9

rusty-key:
Sorry, I am new to the project, so I might be missing something obvious. What you are sharing is an option of passing declarations separately, this works for me. But what I am having problem with, is direct calls to `Css.unsafe`:
```
  (~color: Css.Types.Color.t) => {
    "background: blue";
    [Css.unsafe("color", "trust-me")]
  }
];
```
↓
```
[1/7] Building src/components/Greet.reast
FAILED: src/components/Greet.reast 
...
File "/Users/rustykey/code/exp/try-styled-ppx/src/components/Greet.re", line 6, characters 22-23:
6 |     [Css.unsafe("color", "trust-me")];
                          ^
Error: Parse error while reading token 'EOF'
```
Looks like ppx is trying to parse it and failing, which kinda makes sense. I am not sure I understand how it supposed to work

I am using `1.1.1-nightly.ae39eb6.0` btw, in case there are differences in latest version.

EduardoRFS:
@rusty-key oh yeah, you're right, that's a bug, nice catch, can you try

```reason
let color_string = "color";
let color_value = "trust-me";
  (~color: Css.Types.Color.t) => {
    "background: blue";
    [Css.unsafe(color_string, color_value)]
  }
];
```

But we need a proper solution for this, a couple of options are

- an attribute so that it will ignore the string, like `[@ignore]"color"`
- an named string `{ignore|color|ignore}`
- detecting that is `Css.unsafe` and stops parsing it

I think all of them are quite simple to implement, I can do it myself. We just need a design for it, what do you think @davesnx 

## About the PR

I'm not particularly against this PR, but I would like if it was explicit, like `unsafe-color`. But even then we should fix the current escape hatch.

rusty-key:
> @rusty-key oh yeah, you're right, that's a bug, nice catch, can you try

yeah, that works. Though, I cannot make it work on argument-less components: 
```
module Component = [%styled.div
  "background: blue";
  [Css.unsafe(color_string, color_value)];
];
↓
  We've found a bug for you!
  /Users/rustykey/code/exp/try-styled-ppx/src/components/Greet.re 5:35-44
  
  3 │ let color_string = "color";
  4 │ let color_value = "trust-me";
  5 │ module Component = [%styled.div "background: blue"; [Css.u
      nsafe(color_string, color_value)]];
  6 │ 
```

I guess if that would work, that would be good enough


> detecting that is Css.unsafe and stops parsing it

You mean automatically detect if the property is not supported? If so, it's not exactly what I am looking for, I need manual control to have an escape hatch for certain cases. Any greppable annotation would work for me, and I think that it should be annoying enough so people would not start using it left and right.

davesnx:
Here https://github.com/davesnx/styled-ppx/blob/master/packages/ppx/src/styled_ppx.re#L574-L580 is where we deal with different render strategies for different ppx extension points, renderPayload is called from different mappers. Probably removing `render_emotion_style` and using `Css_to_emotion.render_declaration_list` will deal with unsupported properties globally.

PR #244, opened 2022-01-23
feat: implement selectors parser

joseemds:


packages/ppx/src/css_to_emotion.re
```diff
@@ -212,116 +207,73 @@ and render_style_rule = (ident, rule: Style_rule.t): Parsetree.expression => {
...
+                  | "read-write" => "readWrite"
+                  | c => c
 
+    }
+    | Pseudoelement((e, _)) => switch(e) {
+                  | "first-line" => "firstLine"
+                  | "first-child" => "firstChild"
+                  | "first-letter" => "firstLetter"
+                  | "spelling-error" =>  "spellingError"
+                  | "grammerError" => "grammarError"
```
davesnx, line 236: is gammerError a typo?

packages/ppx/test/bucklescript/src/selectors_test.re
```diff
@@ -19,27 +19,27 @@ let testData = [
...
-          {js|&:nth-child(even)|js},
+          {js|& :nth-child(even)|js},
           [|CssJs.color(CssJs.red)|]
         )
       |]
     ),
   ),
   (
     "nth-child(3n+1)",
-    [%cx "& > div:nth-child(3n+1) { color: blue; }"],
```
davesnx, line 30: Does it mean that it breaks the parser  if spaces aren't provided?

packages/ppx/test/bucklescript/src/selectors_test.re
```diff
@@ -54,7 +54,7 @@ let testData = [
     [%cx "& + & { margin: 10px; }"],
     CssJs.style(. [|
       CssJs.selector(.
-        {js|& + &|js},
```
davesnx, line 57: Maybe we can trim the resultant selector to match perfectly?

packages/ppx/test/bucklescript/src/__snapshots__/array_test.bs.js.snap
```diff
@@ -41,6 +42,7 @@ exports[`SequenceDynamicComponent renders 1`] = `
 <div>
   <div
     class="emotion-0"
+    var="White"
```
davesnx, line 45: Does that seem like an accident? We don't want to render variables into dom attributes (I fixed this last month). Don't worry about it, I will handle it. Try to open an issue if you merge this PR

Edit: This is the cause of failing the CI.

davesnx (approved):
Super happy to have this, will eliminate a bunch of problems and make it selectors much safer to treat.

One thing that it's a little stylish is about running refmt in the .re files and maintaining format in mly files. (We currently don't run refmt since breaks with sedlex)

packages/ppx/src/css_to_emotion.re
joseemds: yes, thanks!

packages/ppx/test/bucklescript/src/selectors_test.re
joseemds: No, this was just me editing the wrong string

packages/ppx/test/bucklescript/src/selectors_test.re
```diff
@@ -19,7 +19,7 @@ let testData = [
     CssJs.style(.
       [|
         CssJs.selector(.
-          {js|&:nth-child(even)|js},
+          {js|& :nth-child(even)|js},
```
davesnx, line 22: I believe this isn't valid right?

packages/ppx/src/css_to_emotion.re
```diff
@@ -168,39 +168,34 @@ and render_style_rule = (ident, rule: Style_rule.t): Parsetree.expression => {
...
 
   let rec render_prelude_value = (s, (value, value_loc)) => {
+
     switch (value) {
     | Delim(":") => ":" ++ s
     | Delim(".") => "." ++ s
     | Delim(",") => ", " ++ s
     /* v can be ">", so we need an empty space between */
-    | Delim(v) => " " ++ v ++ " " ++ s
+    | Delim(v) => v ++ " " ++ s
```
davesnx, line 177: Do we have a  list of valid delimiters? Seems like here we have a few branches but are very poorly. Maybe we could have something similar as get_pseudoelement and have the entire list of them?

This would make the parsing more safe and better on error messaging.

packages/ppx/test/bucklescript/src/selectors_test.re
joseemds: i tested in the demo using:
```
[%styled.div {|
  &:nth-child(even){
    display: none;
  }
|}]
```

that output this, and it works

packages/ppx/src/css_to_emotion.re
joseemds: I agree, but if we have a list of valid delimiters i couldnt find it, i searched in the CSS spec and found this [page](https://www.w3.org/TR/selectors-4/#grammar), but i dont know if all the delimiters are there

Issue #47, opened 2020-03-12
Support CSS functions


davesnx:
Functions are defined in the parser but aren't allowed in any property. Using any of those properties will break:

- calc
- max/min
- minmax
- attr
- blur
- radial-gradient() and repeating-radial-gradient()
- clamp
- repeat (https://github.com/davesnx/styled-ppx/issues/146)

PR #248, opened 2022-02-04
Fix calc

davesnx:
Fixes https://github.com/davesnx/styled-ppx/issues/47

- Organize declarations_to_emotion
- Used calc in `widows`
- Parse calc properly: `dimension` is invalid. That why it was failing at parsing and the issue fixed on Delim.
- Render calc properly: needed to create a bunch of `extended-*` properties where calc can sneak in. Created the system of functions to abstract from this issues on the rendering side.


packages/ppx/src/declarations_to_emotion.re
```diff
@@ -561,20 +587,25 @@ let render_named_color =
...
+  | `Extended_percentage(pct) => render_extended_percentage(pct);
 
 let render_function_rgb = ast => {
-  let to_number = percentage => percentage *. 2.55;
+  let color_to_float = v => render_integer(v |> int_of_float);
+
+  let to_number = fun
+    // TODO: bs-css rgb(float, float, float)
+    | `Percentage(pct) => color_to_float(pct *. 2.55)
+    | _ => render_integer(999);
```
joseemds, line 599: Is this unreachable? If so, isnt better to fail here? So it would be easy to catch some possible error in the future

davesnx: Good catch, was missing a function_calc here!

PR #249, opened 2022-02-04
Support interpolation in length values

rusty-key:
ditto

Allows: 
```reason
[%css {|width: $(value);|}];
```


rusty-key:
@davesnx, please take a look, this is the first step with `length`. If it's ok, I'll continue with other values. 

Unfortunately, I stumbled upon `border-radius`' type, trying to figure it out.

davesnx:
Closed in favor of 253

PR #255, opened 2022-03-07
Add support to interpolation in media query

davesnx:


packages/ppx/test/snapshot/test.re
```diff
@@ -49,9 +49,6 @@ module StringInterpolation = [%styled.div {j|
 let className = [%cx "display: block;"];
 let classNameWithMultiLine = [%cx {| display: block; |}];
 
-let classNameWithArray = [%cx [| cssProperty |]];
-let cssRule = [%css "color: blue;"];
-let classNameWithCss = [%cx [| cssRule, [%css "background-color: green;"] |]];
```
davesnx, line 54: Any reason why these are deleted?

packages/ppx/src/declarations_to_string.re
```diff
@@ -0,0 +1,425 @@
...
+    )
+    |> String.concat("");
+
+  /* Transform property name to camelCase since we bind emotion to the Object API */
+  let propertyName = property |> to_camel_case |> render_string;
+  let value = value |> render_string;
+
+  %expr
+  "(" ++ [%e propertyName] ++ ": " ++ [%e value] ++ ")";
+};
```
davesnx, line 393: Do we need the rendering unsafe here? I think we handle all cases nicely

packages/ppx/src/declarations_to_string.re
```diff
@@ -0,0 +1,425 @@
...
+  | `Scale_down => id([%expr "scale-down"])
+  | `Cover => id([%expr "cover"])
+  | `Full_width => raise(Unsupported_feature)
+  | `Unset => id([%expr "unset"])
+  | `Full_size_kana => raise(Unsupported_feature);
+
+let render_number = (number, unit) =>
+  string_of_int(Float.to_int(number)) ++ unit |> render_string;
+let render_percentage = percentage =>
+  "%" ++ string_of_float(percentage *. 100.) |> render_string;
```
davesnx, line 104: Is that correct? seems like the wrong order

packages/ppx/src/declarations_to_string.re
```diff
@@ -0,0 +1,425 @@
...
+  let ast_to_expr = ast => value_of_ast(ast) |> value_to_expr;
+  let string_to_expr = string =>
+    ast_of_string(string) |> Result.map(ast_to_expr);
+
+  {ast_of_string, value_of_ast, value_to_expr, ast_to_expr, string_to_expr};
+};
+
+let variants_to_string =
+  fun
+  | `Row => id([%expr "row"])
```
davesnx, line 42: A little random idea: What about generating this render methods from the Parser.re ppx? I think there we have all the information needed to turn from variants/combinations of values into strings already. Nothing that we want to do now, but maybe worth considering? WDYT?

packages/ppx/src/css_to_emotion.re
```diff
@@ -90,23 +95,24 @@ and render_media_query = (ar: At_rule.t): Parsetree.expression => {
...
+          switch (Declarations_to_string.parse_declarations(ident, value)) {
           | Error(`Not_found) =>
             grammar_error(ident_loc, "unsupported property: " ++ ident)
           | Error(`Invalid_value(_error)) =>
             grammar_error(loc, "invalid value")
-          | Ok(_) => ()
+          | Ok(exprs) => exprs
           };
-        source_code_of_loc(complete_loc);
+          List.fold_left(concat(~loc), acc, exprs);
```
davesnx, line 106: Sweet! 

Do we handle the parsing errors before?

davesnx (commented):
Great job, maybe adding some tests on Static or Typecheck would make the implementation more robust.

packages/ppx/test/snapshot/test.re
joseemds: It was an accident, i will add it again

packages/ppx/src/declarations_to_string.re
joseemds: There are some properties that weren't supported, like some length units, but since this case is just strings and this function only escape the string value, i think its reasonable to handle everything and remove this.

packages/ppx/src/declarations_to_string.re
joseemds: Right, thanks!

packages/ppx/src/declarations_to_string.re
joseemds: I think that it makes a lot of sense, it would be easier to just add more properties and it would remove the necessity to write a lot of  similar functions to `variant_to_string`

packages/ppx/test/bucklescript/src/media_query_test.re
```diff
@@ -48,7 +48,7 @@ let testData = [
            and (orientation: portrait) { color: brown; }"],
     CssJs.style(. [|
       CssJs.media(.
-        "all and (max-width: 218px) and (max-height: 281px) and (orientation: portrait)",
+        "all and (max-width: 218px) and (max-height: 281px) and (orientation: portrait) ",
```
davesnx, line 51: Maybe we should String.trim the selector?

joseemds: I tried to apply `String.trim`, but different from the string concatenation, bucklescript/rescript doesnt inline the result, so the output would be more verbose (`CssJs.media($$String.trim("all and (max-width: 218px) and (max-height: 281px) and (orientation: portrait) ")`) and instead of adding a whitespace we would add `String.trim` to all media tests.

Do you prefer to change it or have other suggestion on how to do it?

davesnx: It's not really critical to be honest, but looks dirty. Let's not block this PR because of this, we can revisit this.

packages/ppx/src/css_to_emotion.re
joseemds: `Declarations_to_string` will raise if the value is invalid or the property is not found, and i just commited a fix to raise when the prelude is empty, preventing rules like this `@media {color: blue}` to be valid

PR #257, opened 2022-04-01
test: improve selectors suite

joseemds:
The tests that are commented out are breaking the parser or doesn't output correctly.

E.g: 
```
p:first {} // breaks.
&:nth-child(even of li, .item) {color: blue} // Doesn't break, but the output is wrong (&:nth-child(evenofli, .item))

```


davesnx:
Opened a PR migrating those tests into native: https://github.com/davesnx/styled-ppx/pull/259

PR #262, opened 2022-04-12
Support transform function

joseemds:


davesnx (approved):
Great job!

packages/reason-css-parser/lib/Parser.re
```diff
@@ -154,7 +154,7 @@ and function_minmax = [%value.rec "minmax( [ <extended-length> | <extended-perce
 and function_opacity = [%value.rec "opacity( <number-percentage> )"]
 and function_paint = [%value.rec "paint( <ident> ',' [ <declaration-value> ]? )"]
 and function_path = [%value.rec "path( <string> )"]
-and function_perspective = [%value.rec "perspective( <extended-length> )"]
+and function_perspective = [%value.rec "perspective( <property-perspective> )"]
```
davesnx, line 157: The extended-length adds the possibility to interpolate, is it the case for property-perspective?

joseemds: Yes, `property-perspective` is defined as: `and property_perspective = [%value.rec "'none' | <extended-length>"]`, i made this change because `property-perspective` and `function-perspective` share the same spec: 

https://drafts.csswg.org/css-transforms-2/#funcdef-perspective
https://drafts.csswg.org/css-transforms-2/#propdef-perspective

davesnx: I see, I made the extended prefix to exemplify that it's a property that is modified from the spec. Would it make sense to have extended_property_perspective? Just overthinking the naming to be honest.

joseemds: I think its valid change, but is it necessary? since we aim to support interpolation in all properties, in the end everything would have the `extended_` prefix

davesnx: True that

PR #275, opened 2022-06-03
VIM highlighting

ixzzd:
In this PR I am working on highlighting for vim.
The idea is to highlight `%cx` and `%styled` ppx payload with `syntax/css.vim` syntax rules

### DIFF
--- a/packages/editors/vim/after/syntax/ocaml/styled.vim
+++ b/packages/editors/vim/after/syntax/ocaml/styled.vim
@@ -1,0 +1,19 @@
+if polyglot#init#is_disabled(expand('<sfile>:p'), 'css', 'after/syntax/ocaml/styled.vim')
+  finish
+endif
+
+if exists('b:current_syntax')
+  let s:current_syntax = b:current_syntax
+  unlet b:current_syntax
+endif
+
+let b:styled_nested_syntax = 1
+syn include @CssSyntax syntax/css.vim
+unlet b:styled_nested_syntax
+
+if exists('s:current_syntax')
+  let b:current_syntax = s:current_syntax
+endif
+
+syntax region styledPpxEncl matchgroup=ocamlPpxEncl start=+\[%cx+ end=+\]+ keepend contains=styledBody
+syntax region styledBody start=+{|+ end=+|}+ contains=@CssSyntax,styledPpxInterpolation contained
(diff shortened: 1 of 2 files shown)


davesnx:
Hey @ixzzd, Thanks for working on this!

Regarding your doubts on "Better interpolation". Interpolation is between `$(` and `)` and currently are the same as 
OCaml idents plus DOT (so we can have `Module.name`, `name`, etc), in any case should look like: `{ (A…Z) ∣ (a…z) ∣ 0…9 ∣ _ ∣ . }` (eventually will support `'` as well).

packages/editors/vim/after/syntax/ocaml/styled.vim
```diff
@@ -0,0 +1,20 @@
...
+syn include @CssSyntax syntax/css.vim
+unlet b:styled_nested_syntax
+
+if exists('s:current_syntax')
+  let b:current_syntax = s:current_syntax
+endif
+
+syntax region styledPpxEncl matchgroup=ocamlPpxEncl start=+\[%cx+ end=+\]+ keepend contains=styledBody
+syntax region styledPpxEncl matchgroup=ocamlPpxEncl start=+\[%styled+ end=+\]+ keepend contains=styledBody
+syntax region styledBody start=+{|+ end=+|}+ contains=@CssSyntax contained
```
davesnx, line 20: Start can either be `{|` or `"`. It's a regular OCaml string, which doesn't matter what's the termination.

packages/editors/vim/after/syntax/ocaml/styled.vim
```diff
@@ -0,0 +1,20 @@
...
+  finish
+endif
+
+if exists('b:current_syntax')
+  let s:current_syntax = b:current_syntax
+  unlet b:current_syntax
+endif
+
+let b:styled_nested_syntax = 1
+syn include @CssSyntax syntax/css.vim
```
davesnx, line 11: Is there any way we can extend syntax/css.vim and add the interpolation? 
If it's not, does it sound a good idea to fork it? :S

packages/editors/vim/after/syntax/ocaml/styled.vim
ixzzd: Updated in the https://github.com/ahrefs/vim-styled-ppx

packages/editors/vim/after/syntax/ocaml/styled.vim
ixzzd: Yes, I think it does make sense to include interpolation to the CSS highlighting rules.
Added here: https://github.com/ahrefs/vim-styled-ppx/blob/main/syntax/styled-ppx-payload.vim#L524-L525

ixzzd:
@davesnx I updated the PR- moved vim plugin to the separate repo to support vim plugin managers and added https://github.com/ahrefs/vim-styled-ppx link to the documentation

PR #280, opened 2022-06-28
feat: add support to WS token in css_parser.mly

joseemds:
This PR add `WS` token into `css_parser` grammar so we can handle cases that are sensitive to it

---

EDIT:

- Add WS as token (@joseemds)
- Treat all symbols with_whitespace when it doesnt matter at parsing (@joseemds)
- Add DOUBLE_COLON as token, to disambugiate with COLON; COLON;
- Explicit handle AT_MEDIA and AT_KEYFRAMES (the only at_rules that we support)
- Handle empty media
- Handle `@media (color)` syntax
- Transform pseudoclass functions (`:not()`) to handle recursive, but render as strings.
- Change all strings to quoted_strings="js" and uncomment test that relied on raw strings with raw.literal
- Create a lexer-renderer


packages/parser/css_parser.mly
```diff
@@ -46,42 +47,51 @@ stylesheet:
   ;
 
 stylesheet_without_eof:
-  rs = list(rule) { (rs, Lex_buffer.make_loc $startpos $endpos) }
+  rs = list(with_whitespace(rule)) { (rs, Lex_buffer.make_loc $startpos $endpos) } // with whitespace also add a lot of warnings here
```
davesnx, line 50: > with whitespace also add a lot of warnings here

Don't fully understand this comment

packages/parser/css_parser.mly
```diff
@@ -46,42 +47,51 @@ stylesheet:
...
+  | ds = with_loc(declarations); EOF { ds }
   ;
 
 rule:
   | r = at_rule { Rule.At_rule r }
   | r = style_rule { Rule.Style_rule r }
   ;
 
+with_whitespace(X): 
+  | xs = delimited(WS?, X, WS?);  { xs }
```
davesnx, line 64: Does this handle more than one WS?

packages/parser/css_parser.mly
```diff
@@ -46,42 +47,51 @@ stylesheet:
...
-  | name = AT_RULE_WITHOUT_BODY; xs = prelude_with_loc; SEMI_COLON {
+  | name = AT_RULE_WITHOUT_BODY; WS?; xs = with_loc(prelude); SEMI_COLON; {
       { At_rule.name = (name, Lex_buffer.make_loc $startpos(name) $endpos(name));
         prelude = xs;
         block = Brace_block.Empty;
         loc = Lex_buffer.make_loc $startpos $endpos;
       }
     }
-  | name = NESTED_AT_RULE; xs = prelude_with_loc; LEFT_BRACE; s = stylesheet_without_eof; RIGHT_BRACE {
+  | name = NESTED_AT_RULE; WS?; xs = with_loc(prelude); s=brace_block(stylesheet_without_eof) SEMI_COLON; {
```
davesnx, line 80: This seems like a "breaking change". semi_colon become mandatory here?

packages/parser/css_parser.mly
joseemds: I meant menhir warnings, it increased the number of shitf/reduce errors a lot

PR #283, opened 2022-07-27
feat: add inline type annotation on codegen

joseemds:

PR #285, opened 2022-07-28
feat: add inline type annotation on codegen when generating code with interpolation

joseemds:
Alternative to #283

Since notating types everywhere adds a lot of noise, this only add type notation when there is an interpolation.

This happens on `transform_with_variable`, so the properties that call `emit_shorthand` directly and handle `Variable` manually don't have type notation

### DIFF
--- a/packages/ppx/src/declarations_to_emotion.re
+++ b/packages/ppx/src/declarations_to_emotion.re
@@ -22,6 +22,12 @@ type transform('ast, 'value) = {
   string_to_expr: string => result(list(Parsetree.expression), string),
 };
 
+
+let add_type = (expr) => {
+  let typ = Helper.Typ.constr(~loc=Location.none, {txt: Ldot(Lident("CssJs"), "rule"), loc: Location.none}, []);
+  Helper.Exp.constraint_(~loc=Location.none, expr, typ);
+};
+
 let emit = (property, value_of_ast, value_to_expr) => {
   let ast_of_string = Parser.parse(property);
   let ast_to_expr = ast => value_of_ast(ast) |> value_to_expr;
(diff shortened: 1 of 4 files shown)


davesnx (approved):
That's a sweet approach 👍🏼

packages/ppx/src/declarations_to_emotion.re
```diff
@@ -270,7 +276,10 @@ let transform_with_variable = (parser, mapper, value_to_expr) =>
     fun
     | `Variable(name) => render_variable(name)
     | `Value(ast) => mapper(ast),
-    value_to_expr,
+    fun
+    | {pexp_desc: Pexp_ident({txt: Ldot(Lident("CssJs"), _), _}), _} as exp => value_to_expr(exp) // we dont want to add types on cases with `CssJs.red` for example
```
davesnx, line 280: I don't fully understand the comment, or why we are branching for idents.

joseemds: `transform_with_variables` is called from `apply(Parser.spec, [% CssJs.fn], map_fn)` and it generates a function call `CssJs.fn(map_fn(parsed_spec))`.

The `map_fn` usually returns polymorphic variants, so when it is an `Pexp_ident` its because it is using a local value or a value from some module, this would be interpolation, except when it is some value directly from `CssJs`, for example `CssJs.red, CssJs.blue`, since we dont want to add type notation into those, we have the branch where if the module is `CssJs` we just call `value_to_expr` without adding types.

davesnx: Right, so you only want to render the type when `Variable is there?

joseemds: Yes, since `Variable calls `render_variable` that returns a `Pexp_ident`.

Matching `Variable` directly in `emit` didn't worked because the typechecker would complain that `Variable` polyvar didn't exists in some types generated by spec parser.

Doing it in the mapping function would not work too because it would type the arg instead of the function

PR #301, opened 2022-08-11
Create a test file with the draft spec

davesnx:


packages/reason-css-parser/test/Spec.re
```diff
@@ -0,0 +1,99 @@
...
+  | `fit_content
+  | `fit_content_length(length)
+  | `fit_content_percentage(percentage)
+];
+
+/* and final_bg_layer = [%value.rec "<'background-color'> || <bg-image> || <bg-position> [ '/' <bg-size> ]? || <repeat-style> || <attachment> || <box> || <box>"] */
+
+type xor('a, 'b) = [ | `A('a) | `B('b) ];
+type xor3('a, 'b, 'c) = [ | `A('a) | `B('b) | `C('c)];
+type xor7('a, 'b, 'c, 'd, 'e, 'f, 'g) = [ | `A('a) | `B('b) | `C('c) | `D('d) | `E('e) | `F('f) | `G('g) ];
```
joseemds, line 32: Hey, shouldn't this be `or`? 
https://developer.mozilla.org/pt-BR/docs/Web/CSS/Value_definition_syntax#double_bar

joseemds: As the link above describe all the entities are optionals but at least one must be present, how would we encode that? `option` + raise if everything is `None`?

davesnx: Totally, that's an or.

We can't ensure it via the types right?

PR #304, opened 2022-08-24
Add a type generator from CSS Spec

joseemds:
This is not complete, right now it is only running at the `spec.re` file, but it also works at `Parser.re`.

This produces the right output based on #301 and produces an opaque type as a fallback for the non implemented part of the spec.

The snapshot test is not outputting the types, but when i check it with `esy dune describe pp packages/reason-css-parser/test/spec.re.ml` the output is right, i don't know if it is because `Ppxlib.Driver.standalone()` doesn't run `preprocess_impl` or something else and i could not discover the reason.

I want to add some unit tests for `EmitType` but i think that first it should be exposed as a library, so should it live at `reason-css-parser/lib` or somewhere else?

One alternative for not depend on the file path would be to use a command line argument (like `-gen-type`) and pass it only to `Parser.re` and `spec.re` using dune [`per_module` stanza](https://dune.readthedocs.io/en/stable/concepts.html#per-module-preprocessing-specification) WDYT?


packages/reason-css-parser/ppx/Reason_css_parser_ppx.re
```diff
@@ -59,4 +59,26 @@ let valueRecExtension =
     expander(~recursive=true),
   );
 
-Driver.register_transformation(~extensions=[valueExtension, valueRecExtension], "css-value-parser-ppx");
+let gen_type = (str) => {
+  let bindings = List.find_opt(fun
+    | {pstr_desc: Pstr_value(Recursive, _), pstr_loc: loc} => {
+      let filename = loc.loc_start.Lexing.pos_fname;
+      filename == "packages/reason-css-parser/test/spec.re";
```
davesnx, line 66: Ppx doesn't run on Tests because of this

packages/reason-css-parser/ppx/Reason_css_parser_ppx.re
```diff
@@ -59,4 +59,26 @@ let valueRecExtension =
...
+    | Some({pstr_desc: Pstr_value(Recursive, value_bindings), _}) => value_bindings
+    | _ => []
+  }
+
+  if(value_bindings == []){
+    str
+  } else {
+    let types_modules = EmitType.gen_types(value_bindings);
+    str @ [types_modules]
+  }
```
davesnx, line 81: Why not joining the switch?

packages/reason-css-parser/ppx/EmitType.re
```diff
@@ -0,0 +1,269 @@
...
+          };
+          module Ast_builder = Ppxlib.Ast_builder.Make(Loc);
+          module Emit = Make(Ast_builder);
+          try(Emit.create_value_parser(name, ast)) {
+          | Unsupported => Emit.abstract_type(name)
+          };
+        | None => failwith("Error while parsing CSS spec")
+        };
+      inner(rest, [type_, ...acc]);
+    };
```
davesnx, line 261: Why the recursivity? Seems like you are mapping over the value_bindings.

packages/reason-css-parser/ppx/EmitType.re
```diff
@@ -0,0 +1,269 @@
...
+      let type_ =
+        switch (Css_spec_parser.value_of_string(payload)) {
+        | Some(ast) =>
+          module Loc: {let loc: Location.t;} = {
+            let loc = loc;
+          };
+          module Ast_builder = Ppxlib.Ast_builder.Make(Loc);
+          module Emit = Make(Ast_builder);
+          try(Emit.create_value_parser(name, ast)) {
+          | Unsupported => Emit.abstract_type(name)
```
davesnx, line 256: Instead of rising Unsupported, why not run Emit.abstract_type inside create_value_parser?

packages/reason-css-parser/ppx/EmitType.re
```diff
@@ -0,0 +1,269 @@
...
+        ~manifest=None,
+        ~kind=Ptype_abstract,
+      );
+    pstr_type(Recursive, [type_]);
+  };
+  // TODO: multiplier name
+  let rec variant_name = value => {
+    let value_name =
+      switch (value) {
+      | Terminal(Delim(name), _) => value_of_delimiter(name)
```
davesnx, line 94: Would be perfect to annotate the parsing equivalent of Terminal, Group and Combinator.

davesnx (commented):
Left a few comments, but the only that isn't stylish is the hardcoded  path. Fix this and would be cool to have the output of the types on the snapshot

packages/reason-css-parser/ppx/EmitType.re
joseemds: Some places expect a `row_field` instead of a `type_declaration` for generating a variant type, so in more complex cases where we support some features but others not (e.g `Xor` and `And`) i thought it would be simpler and make more sense to raise and catch later returning the abstract type

packages/reason-css-parser/ppx/EmitType.re
joseemds: Can you give an example? I could not see clearly what you mean

packages/reason-css-parser/ppx/Reason_css_parser_ppx.re
joseemds: I removed it, but do we want to run this in all files that contains `the  ppx? Or only in specific files? If its the second option, what do you suggest to achieve this? I think the dune `per_module` + an specific flag (`--gen-types` ?) would be nice, WDYT?

davesnx: For now reason-css-parser's ppx are meant to be running on the same single file. I wouldn't worry too much, since there aren't any other invocations of this ppx elsewhere.

packages/reason-css-parser/ppx/EmitType.re
```diff
@@ -0,0 +1,301 @@
...
+       )
+    |> List.map(
+         fun
+         | (value, None) => value
+         | (value, Some(int)) => value ++ "_" ++ string_of_int(int),
+       )
+    |> List.rev;
+  };
+
+  let mk_typ = (name, types) => {
```
davesnx, line 130: Could you expand `mk_typ` to something a little more expressive? 

makeType or make_type or make_type_assertion or whatever

packages/reason-css-parser/ppx/EmitType.re
```diff
@@ -0,0 +1,301 @@
...
+        ~name=txt(name),
+        ~params=[],
+        ~cstrs=[],
+        ~kind=Ptype_abstract,
+        ~private_=Public,
+        ~manifest=Some(core_type),
+      );
+  };
+
+  let mk_branch = (name, constructor, types) => {
```
davesnx, line 142: Can you make it more expressive as well?

packages/reason-css-parser/ppx/EmitType.re
```diff
@@ -0,0 +1,301 @@
...
+    abstract_type("length"),
+    abstract_type("angle"),
+    abstract_type("time"),
+    abstract_type("frequency"),
+    abstract_type("resolution"),
+    abstract_type("percentage"),
+    abstract_type("ident"),
+    abstract_type("custom_ident"),
+    abstract_type("any_value"),
+    // abstract_type("string"), already represented by OCaml string type
```
davesnx, line 258: This will need to be a real definition, right?

packages/reason-css-parser/test/Types_test.re
```diff
@@ -0,0 +1,81 @@
...
+    gen_type("size", [%expr [%value "relative | static | absolute"]]),
+    [%stri type size = [ | `Relative | `Static | `Absolute]],
+  ),
+  // And
+  (
+    gen_type(
+      "text_emphasis_position",
+      [%expr [%value "[ 'over' | 'under' ] && [ 'right' | 'left' ]"]],
+    ),
+    [%stri type text_emphasis_position],
```
davesnx, line 44: Can you comment the test that aren't "implemented"?

packages/reason-css-parser/test/Types_test.re
```diff
@@ -0,0 +1,81 @@
...
+  (
+    gen_type("contradiction", [%expr [%value "'not' <string>"]]),
+    [%stri type contradiction],
+  ),
+  // Group
+  (
+    gen_type("unsupported", [%expr [%value "test | [not 'supported']"]]),
+    [%stri type unsupported],
+  ),
+  // Function_call
```
davesnx, line 66: I love comments you wrote on the test cases. Would be really cool to update the parser with those names to reflect better.

packages/reason-css-parser/ppx/EmitType.re
```diff
@@ -0,0 +1,301 @@
...
+           let value = total == 1 ? (value, None) : (value, Some(current));
+           [value, ...acc];
+         },
+         [],
+       )
+    |> List.map(
+         fun
+         | (value, None) => value
+         | (value, Some(int)) => value ++ "_" ++ string_of_int(int),
+       )
```
davesnx, line 126: Why fold_left and later map?

packages/reason-css-parser/ppx/EmitType.re
```diff
@@ -0,0 +1,301 @@
...
+let extract_variable_name = (pat: Parsetree.pattern) => {
+  switch (pat.ppat_desc) {
+  | Ppat_var({txt, _}) => txt
+  | _ => failwith("expected variable name")
+  };
+};
+
+let gen_type = (binding: Parsetree.value_binding) => {
+  let name = extract_variable_name(binding.pvb_pat);
+  let (payload, loc) = extract_ppx_content(binding.pvb_expr);
```
davesnx, line 280: These 2 functions can raise and there's seems to no be any reason to not have the switch expressions together inlined here. Thoguhts?

packages/reason-css-parser/test/Types_test.re
```diff
@@ -0,0 +1,81 @@
+open Setup;
+open Ast_helper;
+let loc = Location.none;
+
+let compare = (input, expected, {expect, _}) => {
+  let typ = Str.type_(~loc, Recursive, [input]);
```
davesnx, line 6: Adding the Str.type_ on the test seem wrong. This should be part of the code-gen, right?

davesnx (changes requested):
Seems like it's in a very good shape. There're a few comments that must be addressed, looking great thought.

packages/reason-css-parser/test/Types_test.re
joseemds: `EmitType.gen_types` turn all `type_declarations` into `Str.type_`, i made `EmitType.gen_type` to make unit test easier

packages/reason-css-parser/ppx/EmitType.re
joseemds: I did not check this code, just copied from `EmitPatch`

packages/reason-css-parser/ppx/EmitType.re
```diff
@@ -0,0 +1,301 @@
...
+      | Terminal(Property_type(name), _) =>
+        property_value_name(name) |> value_name_of_css
+      | Group(value, _) => variant_name(value)
+      | Combinator(Static, _) => "static"
+      | Combinator(And, _) => "and"
+      | Combinator(Or, _) => "or"
+      | Combinator(Xor, _) => "xor"
+      | Function_call(name, _) => value_name_of_css(name)
+      };
+    value_name |> first_uppercase;
```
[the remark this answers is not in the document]
joseemds: > Note on polymorphic variant tags: the current implementation accepts lowercase variant tags in addition to capitalized variant tags, but we suggest you avoid lowercase variant tags for portability and compatibility with future OCaml versions.

https://v2.ocaml.org/manual/names.html

packages/reason-css-parser/ppx/EmitType.re
joseemds: I think that the inlined version is not very expressive.
```reason
  let (name, payload, loc) =
    switch (binding) {
    | {
        pvb_pat: {ppat_desc: Ppat_var({txt, _}), _},
        pvb_expr: {
          pexp_desc:
            Pexp_extension((
              _,
              PStr([
                {
                  pstr_desc:
                    Pstr_eval(
                      {
                        pexp_desc:
                          Pexp_constant(Pconst_string(value, loc, _delim)),
                        _,
                      },
                      _attrs,
                    ),
                  _,
                },
              ]),
            )),
          _,
        },
        _,
      } => (
        txt,
        value,
        loc,
      )
    | _ => failwith("Error when extracting CSS spec content")
    }
```

packages/reason-css-parser/ppx/EmitType.re
joseemds: I think so, added this to make the code compile, i think that the simplest solution is make the types using metaquot, e.g:

```reason
[%stri type angle = [
  | `Deg(float)
  | `Grad(float)
  | `Rad(float)
  | `Turn(float)
]]
```

packages/reason-css-parser/test/Types_test.re
joseemds: Can you specify what parser or file are you referring to?

packages/reason-css-parser/ppx/EmitType.re
davesnx: It's very common to have this kind of nested statement on ppxes.

packages/reason-css-parser/ppx/EmitType.re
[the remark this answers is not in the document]
davesnx: Cool

packages/reason-css-parser/test/Types_test.re
davesnx: The css-spec-parser ofcourse

davesnx: Very similar comment I made here: https://github.com/davesnx/styled-ppx/pull/304#discussion_r954552338

packages/reason-css-parser/ppx/EmitType.re
joseemds: fair, this code is simple, after mapping the value to it respective variant name, it checks if there is more than one occurrence, if so, the map after the fold will add a number after the name, it is used to represent polymorphism in some functions like rgb:

```
and function_rgb = [%value.rec "
    rgb( [ <extended-percentage> ]{3} [ '/' <alpha-value> ]? )
  | rgb( [ <number> ]{3} [ '/' <alpha-value> ]? )
  | rgb( [ <extended-percentage> ]#{3} [ ',' <alpha-value> ]? )
  | rgb( [ <number> ]#{3} [ ',' <alpha-value> ]? )
"]
```

where all of the definitions above would generate `Function_rgb` at `variant_name`, but their parameters have different types.

other case is `calc_sum` that can have two different operators, 
```
and calc_sum = [%value.rec "<calc-product> [ [ '+' | '-' ] <calc-product> ]*"]
```

Those operators would be `Static_0` and `Static_1`

davesnx (approved):

PR #307, opened 2022-09-19
Add opam

jchavarri:
- Adds opam file for `styled-ppx`. 
- Removes other packages from the repo and put all libs under single package, for easier pinning & usage
- Rely does not support opam :( so the tests are disabled by renaming `dune` files to `_dune`

Most probably shouldn't be merged until opam support is added to rely, and the tests can be re-enabled.


davesnx:
This PR got super-seeded by https://github.com/davesnx/styled-ppx/pull/318, https://github.com/davesnx/styled-ppx/pull/317, and https://github.com/davesnx/styled-ppx/pull/316.

Since we depend on this. I'm not going to close it, to have it easier to grap the commit hash

davesnx:
opam has been resolved, isn't published but you can pin any commit message after 0.33.0

PR #340, opened 2023-04-12
Add basic renderers

davesnx:


packages/reason-css-parser/ppx/Generate.re
```diff
@@ -636,5 +658,175 @@ module Make = (Ast_builder: Ppxlib.Ast_builder.S) => {
...
+              | `Rem(n) => build_variant(~loc, "em", Some(render_number(~loc, n)))
+              | `Rlh(n) => build_variant(~loc, "rlh", Some(render_number(~loc, n)))
+              | `Vb(n) => build_variant(~loc, "vb", Some(render_number(~loc, n)))
+              | `Vh(n) => build_variant(~loc, "vh", Some(render_number(~loc, n)))
+              | `Vi(n) => build_variant(~loc, "vi", Some(render_number(~loc, n)))
+              | `Vmax(n) => build_variant(~loc, "vmax", Some(render_number(~loc, n)))
+              | `Vmin(n) => build_variant(~loc, "vmin", Some(render_number(~loc, n)))
+              | `Vw(n) => build_variant(~loc, "vw", Some(render_number(~loc, n)))
+              | `Zero => build_variant(~loc, "zero", None);
+            }],
```
davesnx, line 819: Why those standard printers are different from the previous implementation?

packages/reason-css-parser/ppx/Generate.re
```diff
@@ -636,5 +658,175 @@ module Make = (Ast_builder: Ppxlib.Ast_builder.S) => {
...
+	}
+
+	and variant_of_values = (kind, values) => {
+		let string_of_kind = fun
+		| Xor => "Xor"
+		| Static => "Static"
+		| Or => "Or"
+		| And => "And"
+		
+		let args = List.mapi((i, _) => Format.sprintf("arg_%d", i) |> txt |> ppat_var, values);
```
davesnx, line 723: instead of mapping thought args, might be worth trying the approach I mention and use the name for a render_* that already exists (or to try to create a render)

davesnx (commented):
2 things:

- Can you run the formatted on Generate.re?
- Would you add a few test cases for the renderers as well?

packages/reason-css-parser/ppx/Generate.re
joseemds: Because we dont have access to the variants parameter `n`; 

before we would do like this:

```\`Rem(n) -> [%expr \`rem([%e render_number(~loc, n))]```

where n would be parsed before being wrapped into a `pexp_variant`;

`build_variant` is a function that is being generated to be called by the generated code so we can achieve the same as before

`let build_variant = (~loc, name, args) => Ast_helper.Exp.variant(~loc, name, args) `

So that the same process occurs (render the arguments and then wrap it into a `pexp_variant`

Comment (davesnx) [reply packages/reason-css-parser/ppx/Generate.re:819]: Right, I prefer to remove the build_variant function and keep them as close as before.
```
| `Zero => [%expr `zero]
| `Vmin(n) => [%expr `vmin([%e render_number(~loc, n)]]
```

PR #346, opened 2023-05-23
Add test to show missing error locations when applying styled-ppx.lib

jchavarri:
Adds a cram test to show that locations are missing when syntactic errors happen when using `styled-ppx.lib`.


davesnx:
Fixed here: https://github.com/davesnx/styled-ppx/pull/347#issuecomment-1560964858

I didn't checkout your branch since it was in a fork and you are collaboration, so feel free to push it to origin <3

PR #364, opened 2023-10-24
add server-rr to deps

jchavarri:


davesnx:
That's not required since it's pinned and the opam worflow isn't polished

jchavarri:
@davesnx This was helpful to avoid issues while installing styled-ppx in ahrefs monorepo, where opam would start installing styled-ppx before the required version of server-rr was available.

> since it's pinned 

The fact that SRR is pinned is orthogonal to having the package added to styled-ppx `depends` field. The suggested fix just helped opam establish the ordering in which packages have to be installed.

davesnx:
Sure it should be added on the opam's monorepo but not in this opam file yet. It won't install here

jchavarri:
I see what you mean. When creating the switch, this addition would be problematic because it would point to a version `dev` of SSR that opam wouldn't know about.

In that case, maybe `pin-depends` should be used rather than manual pinning through Makefile commands? I think it's good to have a version of the opam file of styled-ppx that we can just copy/paste into the monorepo over time (rather than copying the latest, and then try to remember what things we have to add....)

PR #369, opened 2023-11-03
Remove lexbuffer in favor of sedlex

lubegasimon:
**Changes made:**
- removed `Lex_buffer.re` because we no longer have to call from it, instead we rely on `sedlex`-defined functions and a custom
- created a custom`latin1` function


packages/parser/lib/util.re
```diff
@@ -0,0 +1,9 @@
+module Location = Ppxlib.Location;
+
+let make_loc =
+    (~loc_ghost=false, start_pos: Lexing.position, end_pos: Lexing.position)
```
davesnx, line 4: Do you think this file can be called Parser_location and have a make function instead?

packages/parser/lib/Driver_.re
```diff
@@ -0,0 +1,66 @@
+module Lexer = Css_lexer;
```
davesnx, line 1: Any reason why Driver has an underscore?

davesnx (approved):
Looks good to me!

Have you tried updating sedlex to latest?

packages/parser/lib/util.re
lubegasimon: Yes, that is fine assuming that it will only serve that purpose. Otherwise, I named it so, in case we ever need any other helper function that won't be constrained to one use purpose.

packages/parser/lib/Driver_.re
lubegasimon: Yes! To avoid conflicts with `ppxlib`'s Driver module because it is opened in `css_to_emotion` in which we also reference `last_buffer`. When the latter is overwritten with a better implementation, I guess we shall be free to remove the underscore, otherwise, I am okay with using another better name :)

lubegasimon:
> Have you tried updating sedlex to latest?

Yes, in my switch v3.2 is installed  (which is the latest in opam).

davesnx:
Can you add it on the .opam file?

lubegasimon:
> Can you add it on the .opam file?

Done in https://github.com/davesnx/styled-ppx/pull/375

packages/parser/lib/Driver_.re
davesnx: Can we find another name then?

davesnx: Can we fix the ppxlib driver conflict with Driver, then? and keep "Driver.re" as a name here?

lubegasimon: No problem!
How is `parse` after we have renamed [Payload.parse](https://github.com/davesnx/styled-ppx/blob/174da50b8c7201c01554b320202f023a77b20f28/packages/ppx/src/payload.re#L39) to `Payload.parse_payload` or `parse_helper`?

lubegasimon: Yes, done in https://github.com/davesnx/styled-ppx/pull/372/commits/13516ca436eed22a775bbab456a37d91ee352c3e and https://github.com/davesnx/styled-ppx/pull/372/commits/26ec1f6834e2b8e913dc948b863307ffbbc20a00

davesnx (commented):
I would like to merge this, but before that can you add a tests for an error of the parser?

I believe this will be very beneficial while we change the location and container_lum.

packages/parser/lib/Driver_.re
lubegasimon: Right, I will rebase on top of the branch containing the commits

lubegasimon: Done rebasing, meaning https://github.com/davesnx/styled-ppx/pull/372 goes in first instead of  https://github.com/davesnx/styled-ppx/pull/369

packages/parser/lib/util.re
davesnx: Right, I prefer to not have non-ambigious names such as Utils on this codebase, and keep the module focused on one thing

lubegasimon: Done in https://github.com/davesnx/styled-ppx/pull/369/commits/70f8948835d278236b3662361951f2a353c9f2d8

lubegasimon:
> I would like to merge this, but before that can you add a tests for an error of the parser?
> 
> I believe this will be very beneficial while we change the location and container_lum.

I guess this is resolved here https://github.com/davesnx/styled-ppx/pull/369/commits/fcc87416f6604b75d2b905c2ab74e8f995a6b347

packages/parser/test/css_parser_test.re
```diff
@@ -4,14 +4,22 @@ let parse = input => {
...
 let error_tests_data =
   [
-    ("{}", "Parse error while reading token '{'"),
-    ("div { color: red; _ }", "Parse error while reading token '}'"),
+    ("{}", "Parse error while reading token '{' at position 0"),
+    (
+      "div { color: red; _ }",
+      "Parse error while reading token '}' at position 19",
+    ),
+    ("@media $", "Parse error while reading token '$' at position 7"),
```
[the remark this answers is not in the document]
lubegasimon: Fixed in https://github.com/davesnx/styled-ppx/pull/369/commits/bc043cd6138df5e7038779f0873ea80b2c496298

packages/parser/lib/css_lexer.re
```diff
@@ -28,9 +28,7 @@ let unicode_range = [%sedlex.regexp?
   (up_to_6_hex_digits, '-', up_to_6_hex_digits)
 ];
 
-let escape = [%sedlex.regexp?
-  unicode | ('\\', Compl('\r' | '\n' | '\012' | hex_digit))
-];
+let escape = [%sedlex.regexp? '\\'];
```
davesnx, line 31: Why this regex has changed?

packages/parser/lib/css_lexer.re
```diff
@@ -352,44 +350,61 @@ let unreachable = () =>
...
+let consume_escaped = buf => {
+  switch%sedlex (buf) {
+  // TODO: spec typo? No more than 5?
+  | Rep(hex_digit, 1 .. 6) =>
+    let hex_string = "0x" ++ lexeme(buf);
+    let char_code = int_of_string(hex_string);
+    let char = uchar_of_int(char_code);
+    let _ = consume_whitespace(buf);
+    char_code == 0 || is_surrogate(char_code)
+      // U+FFFD is a character used as a substitute for an uninterpretable character from another encoding
```
davesnx, line 373: Where this U+FFFD is being treated?

packages/parser/test/css_parser_test.re
```diff
@@ -4,14 +4,35 @@ let parse = input => {
...
+    let pos = loc.Css_types.loc_start;
+    let curr_pos = pos.pos_cnum;
+    let lnum = pos.pos_lnum;
+    let pos_bol = pos.pos_bol;
+    let err =
+      Printf.sprintf(
+        "%s on line %i at position %i",
+        msg,
+        lnum,
+        curr_pos - pos_bol,
```
davesnx, line 17: Do you think is worth it add this level of detail on the error reporting by the ppx?

davesnx (approved):
Nice

packages/parser/lib/css_lexer.re
lubegasimon: The other parts were redundant and caused issues while starting with escape chars

packages/parser/lib/css_lexer.re
lubegasimon: It is referenced as a string on the next two lines. Though I am not so convinced with it.

packages/parser/lib/css_lexer.re
davesnx: 👍🏼

packages/parser/test/css_parser_test.re
lubegasimon: Yes, I think it is for clarity.

PR #370, opened 2023-11-03
Remove uncesssary tokens

lubegasimon:
**Changes made:**
- removed `CDO` & `CDC` tokens since we ignore comments during lexing via the `discard_comments` function defined in `packages/parser/lib/css_lexer.re`.
- removed the `BAD_STRING` token because the css_parser ensures all strings are correct.
- renamed some tokens.


lubegasimon:
Changes referenced in https://github.com/davesnx/styled-ppx/pull/368

PR #371, opened 2023-11-03
Raise exn if last buffer is empty

lubegasimon:
This takes care of cases where `Fatal error: exception Invalid_argument("index out of bounds")` would be thrown


lubegasimon:
Changes carried in by https://github.com/davesnx/styled-ppx/pull/368

PR #377, opened 2023-11-07
Update hash function to match emotion

jchavarri:
Updates the hash function in native so that it behaves exactly like the original one in emotion.

This has the advantage that there's no need to patch any existing user of styled-ppx & server-reason-react to override the emotion/hash function on the melange build, plus we lower the chances of collision / issues in the hashing algorithm as the original Emotion hash has been tested over time.

The PR also adds some basic cram test to make sure the results returned by each implementation are the same.

@davesnx Feel free to discard or take anything from the PR.

### DIFF
--- a/packages/emotion-hash/hash.ml
+++ b/packages/emotion-hash/hash.ml
@@ -42,34 +42,20 @@ let default str =
     in
 
     let k = add (mul (k land 0xffffl) m) (mul (k lsr 16) 0xe995l lsl 16) in
-    print_endline (Printf.sprintf "K INIT: %lx" k);
     let k = k lxor (k lsr r) in
 
-    print_endline (Printf.sprintf "K FINAL: %lx" k);
     h :=
       add (mul (k land 0xffffl) m) (mul (k lsr 16) 0xe995l lsl 16)
       lxor add (mul (!h land 0xffffl) m) (mul (!h lsr 16) 0xe995l lsl 16);
 
     i := !i + 4;
     len := !len - 4
-    (* print_endline (Printf.sprintf "loop %d" !i);
-       print_endline (Printf.sprintf "len %d" !len);
-       print_endline (Printf.sprintf "H is: %lx" !h) *)
   done;
-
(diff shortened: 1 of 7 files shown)


packages/emotion-hash/hash.ml
```diff
@@ -1,133 +1,72 @@
...
+    h :=
+      add (mul (k land 0xffffl) m) (mul (k lsr 16) 0xe995l lsl 16)
+      lxor add (mul (!h land 0xffffl) m) (mul (!h lsr 16) 0xe995l lsl 16);
+
+    i := !i + 4;
     len := !len - 4
   done;
 
   (* Handle the last few bytes of the input array *)
-  (match !len with
```
jchavarri, line 104: This was problematic. In the original implementation, there's a `switch` with 3 cases, but there's no `break` on each case, so if `len` is 3, all the cases will be executed.

Using `match` doesn't follow that behavior, as only one branch is executed, it'd be analog to JS code having `break` on each case.

I ended up doing a sequence of `if` statements without `else`s, see below.

packages/emotion-hash/hash.ml
```diff
@@ -1,133 +1,72 @@
...
-  hash :=
-    ((!hash & 0xffffl) * 0x5bd1e995l) ++ ((!hash >>> 16) * 0x5bd1e995l << 16);
-  (* (h ^ (h >>> 15)) >>> 0 *)
-  !hash ^ (!hash >>> 15) >>> 0
+  (* Do a few final mixes of the hash to ensure the last few
+     bytes are well-incorporated. *)
+  h := !h lxor (!h lsr 13);
 
-let default (str : string) = str |> murmur2 |> to_css
+  h := add (mul (!h land 0xffffl) m) (mul (!h lsr 16) 0xe995l lsl 16);
```
jchavarri, line 71: idk how to override the `+` and `*` to use the Int32 ones. That'd make this code a bit easier to read.

davesnx:
That's amazing

PR #388, opened 2023-12-04
add support for tab-size property

lubegasimon:


packages/css/native/shared/Css_AtomicTypes.ml
```diff
@@ -122,6 +124,7 @@ module Length = struct
     | `pc x -> Std.Float.toString x ^ {js|pc|js}
     | `pt x -> Std.Int.toString x ^ {js|pt|js}
     | `zero -> {js|0|js}
+    | `num x -> Std.Int.toString x
```
davesnx, line 127: This should be on TabSize module instead of Length

lubegasimon: The current commits are very wrong, I was silly then.

packages/reason-css-parser/lib/Parser.re
```diff
@@ -1569,7 +1569,7 @@ and property_stroke_linejoin = [%value.rec "'miter' | 'round' | 'bevel'"]
 and property_stroke_miterlimit = [%value.rec "<number-one-or-greater>"]
 and property_stroke_opacity = [%value.rec "<alpha-value>"]
 and property_stroke_width = [%value.rec "<svg-length>"]
-and property_tab_size = [%value.rec "<integer> | <extended-length>"]
+and property_tab_size = [%value.rec " <number> | <extended-length>"]
```
davesnx, line 1572: Why change from integer to number?

https://developer.mozilla.org/en-US/docs/Web/CSS/tab-size#values

packages/css/native/shared/Css_AtomicTypes.ml
```diff
@@ -261,6 +261,12 @@ module FontStyle = struct
     | `oblique -> {js|oblique|js}
 end
 
+module TabSize = struct
+  type t = [ `num of float ]
```
davesnx, line 265: `num of int`

packages/ppx/src/declarations_to_emotion.re
```diff
@@ -2093,7 +2093,24 @@ let text_transform =
   );
 let white_space =
   variants(Parser.property_white_space, (~loc) => [%expr CssJs.whiteSpace]);
-let tab_size = unsupportedProperty(Parser.property_tab_size);
+
+let render_tab_size = (~loc, value: Types.property_tab_size) => {
+  switch (value) {
+  | `Number(n) =>
+    int_of_float(n) < 0
+      ? raise(Invalid_value("tab-size value can not be less than 0!"))
```
davesnx, line 2101: This error doesn't happen at code-generation, it happens already at parsing. No need to raise anything/check for it

packages/ppx/test/native/Static_test.re
```diff
@@ -994,7 +994,37 @@ let properties_static_css_tests = [
...
+    [%css "tab-size: calc(10px + 10pt)"],
+    [%expr [%css "tab-size: calc(10px + 10pt)"]],
+    [%expr CssJs.tabSize(`calc(`add((`pxFloat(10.), `pt(10)))))],
+  ),
   // unsupported
+  /* (
+       [%css "tab-size: calc(10 + 10)"],
+       [%expr [%css "tab-size: calc(10 + 10)"]],
+       [%expr CssJs.tabSize(`calc(`add((`num(10), `num(10)))))],
+     ), */
```
davesnx, line 1027: This doesn't have anything to do with tab-size

packages/reason-css-parser/lib/Parser.re
lubegasimon: Formal syntax -- https://developer.mozilla.org/en-US/docs/Web/CSS/tab-size#formal_syntax

packages/ppx/src/declarations_to_emotion.re
lubegasimon: I don't see how it shall error at parsing unless if we have bounds, i.e `... [0,∞]`, which IIRC, is not supported at the moment.

davesnx: If we want to have this kind of validation we would need to create a `let nonnegative-number` on Parser.re (but that's out of the scope from this)

lubegasimon: Right, in that case, do you give this snippet a green light?

davesnx (approved):

PR #391, opened 2023-12-04
add support for line-break

lubegasimon:


packages/ppx/src/declarations_to_emotion.re
```diff
@@ -3348,6 +3350,9 @@ let strokeOpacity =
     render_alpha_value,
   );
 
+let line_break =
+  variants(Parser.property_line_break, (~loc) => [%expr CssJs.lineBreak]);
```
davesnx, line 3354: Maybe adding interpolation into line-break should be nice

lubegasimon:

PR #394, opened 2023-12-04
Support for text-align-last property

lubegasimon:
PR depends on https://github.com/davesnx/styled-ppx/pull/393


lubegasimon:

PR #396, opened 2023-12-04
Support more css props v1

lubegasimon:
**Support for CSS property;**

- text-emphasis-style
- text-emphasis-position
- line-break
- hyphens
- text-align-last
- text-justify
- overflow-inline
- font-synthesis-weight
- font-synthesis-style
- font-synthesis-small-caps
- font-synthesis-position
- font-kerning
- font-variant-position
- font-variant-caps
- font-optical-sizing
- font-variant-emoji
- text-decoration-skip-ink
- text-decoration-skip-box
- text-decoration-skip-inset
- transform-box


packages/css/js/Css_Js_Core.ml
```diff
@@ -645,6 +645,51 @@ let height x =
...
+      | #Var.t as va -> Var.toString va
+      | #Cascading.t as c -> Cascading.toString c )
+
+let textEmphasisPosition x =
+  D
+    ( {js|textEmphasisPosition|js},
+      match x with
+      | #TextEmphasisPosition.OverOrUnder.t as ou ->
+        TextEmphasisPosition.OverOrUnder.toString ou
+      | `textEmphasisPositionYX (y, x) ->
```
davesnx, line 676: I meant another function and the codegen renders one or the other

packages/css/js/Css_Js_Core.ml
```diff
@@ -1045,7 +1123,13 @@ let transformOrigin3d x y z =
       ^ Length.toString z
       ^ {js| |js} )
 
-let transformBox x = D ({js|transformBox|js}, TransformBox.toString x)
+let transformBox x =
+  D
+    ( {js|transform-box|js},
```
davesnx, line 1128: transformBox

packages/css/native/shared/Css_Js_Core.ml
```diff
@@ -1064,7 +1142,12 @@ let transformOrigin3d x y z =
       ^ Length.toString z
       ^ {js| |js} )
 
-let transformBox x = D ({js|transform-box|js}, TransformBox.toString x)
+let transformBox x =
+  D
+    ( {js|transform-box|js},
```
davesnx, line 1147: transformBox

packages/css/js/Css_Js_Core.ml
lubegasimon: I don't think I understand the need for another function, what would it do different?

davesnx: `let textEmphasisPosition`
`let textEmphasisPositions`

So the textEmphasisPosition actual doesn't need to handle the rare `(y, x)` case

lubegasimon: and what shall be the expression, `CssJs.textEmphasisPosition` or `CssJs.textEmphasisPositions`.
This is what I think you meant
```
let textEmphasisPosition x =
  D
    ( {js|textEmphasisPosition|js},
      match x with
      | #TextEmphasisPosition.OverOrUnder.t as ou -> ... )

let textEmphasisPositions value =
  D
    ( {js|textEmphasisPositions|js},
      match value with
      | x, y ->
        (match x with ... (match y with ...))
```
Is that so?

packages/css/js/Css_Js_Core.ml
lubegasimon: Done in https://github.com/davesnx/styled-ppx/pull/396/commits/35494bf086fd0b96418ad6f8d172edd2f5f1350b. Thanks!

packages/css/native/shared/Css_Js_Core.ml
lubegasimon: Done in https://github.com/davesnx/styled-ppx/pull/396/commits/35494bf086fd0b96418ad6f8d172edd2f5f1350b. Thanks!

Issue #359, opened 2023-09-03
Nested selectors can contain issues


davesnx:
Nested selectors should be treated with an empty space at the beginning if they don't start with an `&`.

```css
.$(aaa) .$(bbb) {
  ..
}
// output .aaa .bbb (That's correct 🟢) 

.$(aaa) {
  .$(bbb) {
    ..
  };
}
// output .aaa.bbb (That's wrong 🔴), the output should be same as above (.aaa .bbb)
```

Currently it transform to `CssJs.selector(".aaa", [CssJs.selector(".bbb", [CssJs.color(CssJs.red)])])` so it generates an object to emotion like: `{ ".aaa": {".bbb" : { "color": "red" }} }`

The second selector should have a whitespace at the start

--- 

More test cases:

```
.one_level { &.without_space { color: blue } };

.one_level { & .with_space { color: blue } };

.one_level { &.$(with_interpolation) { color: blue } };

.one_level { & .$(space_with_interpolation) { color: blue } };
```

```
.two_levels { & .first_level { & .second_level { color: blue } } };

...
```

```
.without_ampersand { .one_level { color: blue } }; // expects an space here
```

```
.$(interpolation_at_the_start) { & .one_level { color: blue } };
```

```
#using_ids { & .one_level { color: blue } };
```

```
[using_attrs] { & .one_level { color: blue } };
```

davesnx:
There has been an exploration https://github.com/davesnx/styled-ppx/pull/404 about having a Resolver phase which is the way to go in order to handle the transformation on media-queries and selectors. Similar work has been done at native runtime: packages/emotion/native/Css.ml

lubegasimon:
> Similar work has been done at native runtime: packages/emotion/native/Css.ml

Does it fix the issue (out of curiosity)?

davesnx:
It fixes the same issue but when you aren't using the ppx and running on native

PR #400, opened 2023-12-19
render nested selectors right

lubegasimon:
fixes https://github.com/davesnx/styled-ppx/issues/359

### DIFF
--- a/packages/parser/lib/css_types.re
+++ b/packages/parser/lib/css_types.re
@@ -2,6 +2,8 @@
 
 module Location = Ppxlib.Location;
 
+let nesting_level = ref(0);
+
 [@deriving show({with_path: false})]
 type position =
   Lexing.position = {
@@ -84,9 +86,9 @@ and style_rule = {
 and stylesheet = with_loc(list(rule))
 [@deriving show({with_path: false})]
 and selector =
-  | SimpleSelector(simple_selector)
-  | ComplexSelector(complex_selector)
-  | CompoundSelector(compound_selector)
+  | SimpleSelector(simple_selector, int)
+  | ComplexSelector(complex_selector, int)
+  | CompoundSelector(compound_selector, int)
 [@deriving show({with_path: false})]
 and selector_list = list(with_loc(selector))
 [@deriving show({with_path: false})]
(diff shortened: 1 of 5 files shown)


packages/parser/lib/css_types.re
```diff
@@ -2,6 +2,8 @@
 
 module Location = Ppxlib.Location;
 
+let nesting_level = ref(0);
```
davesnx, line 5: Maybe call it selector_nesting_level?

Probably nice to add a comment explaining what it is also.

packages/parser/lib/driver.re
```diff
@@ -46,7 +46,14 @@ let parse_string = (~skip_whitespace, ~pos, parser, string) => {
 
   last_buffer := Some(from_string(~pos?, string));
 
-  parse(skip_whitespace, buffer, parser);
+  String.contains(string, '}')
+    ? {
+      let () = Css_types.nesting_level := 0;
+      parse(skip_whitespace, buffer, parser);
```
davesnx, line 52: Instead of setting the global on the css_types, I expect to set it to 0 here.

packages/parser/lib/driver.re
```diff
@@ -46,7 +46,14 @@ let parse_string = (~skip_whitespace, ~pos, parser, string) => {
 
   last_buffer := Some(from_string(~pos?, string));
 
-  parse(skip_whitespace, buffer, parser);
+  String.contains(string, '}')
```
davesnx, line 49: This is a really ugly hack, I expect the parse to do this

davesnx (changes requested):

davesnx:
Looking at how to do this during the parsing might be challenging, mostly about when to decrease the nesting_level (The hack of hijacking the parser isn't nice).

I would like to see the approach where the code-gen keeps track of the level, that way we might not need to track a global integer.

lubegasimon:
> Looking at how to do this during the parsing might be challenging, mostly about when to decrease the nesting_level (The hack of hijacking the parser isn't nice).
> 
> I would like to see the approach where the code-gen keeps track of the level, that way we might not need to track a global integer.

Shall we consider this in a separate PR?

PR #401, opened 2023-12-20
store number as float

lubegasimon:


lubegasimon:
It's intentional to store `number` as `string`

PR #402, opened 2023-12-21
refactor code to ease maintenance

lubegasimon:


packages/parser/lib/Tokens.re
```diff
@@ -99,13 +99,18 @@ let token_to_string =
...
   | BAD_URL => "bad url"
-  | BAD_IDENT => "bad indent";
+  | BAD_IDENT => "bad indent"
+  | PERCENTAGE(p) =>
+    "percentage: " ++ string_of_float(p) ++ string_of_char('%')
+  | NUMBER_CSS(n) => "(NUMBER '" ++ string_of_float(n) ++ "')"
+  | DIMENSION_CSS((f, s)) => "dimension: " ++ string_of_float(f) ++ s
+  | AT_KEYWORD(at) => "@ " ++ at;
 
 let token_to_debug =
```
lubegasimon, line 115: I wonder if we could remove `token_to_debug` and instead maintain `token_to_string`

packages/parser/lib/Tokens.re
```diff
@@ -99,13 +99,18 @@ let token_to_string =
...
   | FUNCTION(fn) => fn ++ "("
   | NTH_FUNCTION(fn) => fn ++ "("
   | URL(url) => url ++ "("
   | BAD_URL => "bad url"
-  | BAD_IDENT => "bad indent";
+  | BAD_IDENT => "bad indent"
+  | PERCENTAGE(p) =>
+    "percentage: " ++ string_of_float(p) ++ string_of_char('%')
+  | NUMBER_CSS(n) => "(NUMBER '" ++ string_of_float(n) ++ "')"
+  | DIMENSION_CSS((f, s)) => "dimension: " ++ string_of_float(f) ++ s
```
lubegasimon, line 57: Does it make sense for `NUMBER_CSS` and `DIMENSION_CSS` to be produced as strings as `NUMBER` and `DIMENSION` are stored as strings? In that case, we could remove `NUMBER_CSS` and `DIMENSION_CSS`.  tokens

packages/parser/lib/Tokens.re
```diff
@@ -1,31 +1,3 @@
...
-  | WS // <whitespace-token>
-  | COLON // <colon-token>
-  | SEMI_COLON // <semicolon-token>
-  | COMMA // <comma-token>
-  | LEFT_BRACKET // <[-token>
-  | RIGHT_BRACKET // <]-token>
-  | LEFT_PAREN // <(-token>
-  | RIGHT_PAREN // <)-token>
-  | LEFT_BRACE // <{-token>
-  | RIGHT_BRACE; // <}-token>
```
davesnx, line 25: Are we losing the comments?

lubegasimon: I don't mind because they're self describing

packages/parser/lib/Tokens.re
```diff
@@ -37,38 +9,11 @@ let show_error =
...
-  | RIGHT_BRACKET => "]"
-  | LEFT_PAREN => "("
-  | RIGHT_PAREN => ")"
-  | LEFT_BRACE => "{"
-  | RIGHT_BRACE => "}";
-
-module Parser = Css_parser;
+let string_of_char = c => String.make(1, c);
 
 let token_to_string =
```
davesnx, line 14: This is a bad idea, token_to_string is used for errors which doesn't make any sense to see `IDENT ("foo")` in the error instead of `"foo"`.

I would say to rollback this.

packages/parser/lib/Tokens.re
davesnx: I don't see the difference (either I understand) NUMBER vs NUMBER_CSS.

Why do we need both? Why we need another?

davesnx (changes requested):
Hey Simon, thanks for the effort. I don't think this PR adds any ease of maintenance and degrades the user experience. 

Is there any part of the code that this PR tries to improve that I'm missing? Would you like to share upfront the reasons for the refactors, so I might be able to see the same as you.

lubegasimon:
> Is there any part of the code that this PR tries to improve that I'm missing? Would you like to share upfront the reasons for the refactors, so I might be able to see the same as you.

Well, while unifying the lexers, I duplicated many functions. For example `consume_whitespace_` from `consume_whitespac`, `consume_url_` from `consume_url`. This PR eliminates such duplication and it is a subtask to integrate the `get_next_token` and `consume` functions which I find share more or less the same logic.

packages/parser/lib/Tokens.re
lubegasimon: We could have one if NUMBER_CSS and DIMENSION_CSS are produced as strings

lubegasimon:
sad, but I will rebuild it better in the future when resources are adequate.

PR #405, opened 2024-01-02
Support more css props

lubegasimon:


lubegasimon:
Rebasing ✅

lubegasimon:
A job failed and it's not clear why. This is the second time I am hitting this scenario.
Screenshot 2024-01-20 at 00 35 19

davesnx (changes requested):
let imageRendering and borderImageSource needs to be on Css_Legacy_Core also

PR #406, opened 2024-01-17
Docs clarification

psb:
Minor updates to the docs to fix spelling errors and to add some clarifications. Still some questions:

1) I'm not sure what you are trying to say with [this line](https://github.com/davesnx/styled-ppx/blob/main/packages/website/pages/reference/css.mdx#L17). I have tried to rephrase it as: 

> - Only single properties are allowed; declarations are not allowed. Declarations can contain selectors, but rules can only contain a single pairing of property and value. If you need a selector or more than one rule, use `%cx{:rescript}` instead.

But I'm not sure if this makes sense as I don't really understand the original line.

2) Is there a missing link on [this line](https://github.com/davesnx/styled-ppx/blob/main/packages/website/pages/css-support.mdx#L11)? `[Css bindings]` is in square brackets but I don't know why.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 ### Typed styled components for ReScript
 
-**styled-ppx** is a [ppx](https://tarides.com/blog/2019-05-09-an-introduction-to-ocaml-ppx-ecosystem) that brings styled components to ReScript. Built on top of [emotion](https://emotion.sh), it allows you to style apps safely, quickly, and with zero runtime overhead - just as you have always done it. styled-ppx allows you to create **React Components** with type-safe style definitions that don't rely on a different language ([DSL](https://en.wikipedia.org/wiki/Domain-specific_language)) except CSS.
+**styled-ppx** is a [ppx](https://tarides.com/blog/2019-05-09-an-introduction-to-ocaml-ppx-ecosystem) that brings styled components to ReScript. Built on top of [emotion](https://emotion.sh), it allows you to style apps safely, quickly, and performantly - just as you have always done it. styled-ppx allows you to create **React Components** with type-safe style definitions that don't rely on a different language ([DSL](https://en.wikipedia.org/wiki/Domain-specific_language)) except CSS.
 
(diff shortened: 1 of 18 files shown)


README.md
```diff
@@ -4,7 +4,7 @@
 
 ### Typed styled components for ReScript
 
-**styled-ppx** is the [ppx](https://tarides.com/blog/2019-05-09-an-introduction-to-ocaml-ppx-ecosystem) that brings styled components to ReScript. Build on top of [emotion](https://emotion.sh), it allows you to style apps safe, quickly, performant and as you always done it. Allows you to create **React Components** with type-safe style definitions that don't rely on a different language ([DSL](https://en.wikipedia.org/wiki/Domain-specific_language)) than CSS
+**styled-ppx** is a [ppx](https://tarides.com/blog/2019-05-09-an-introduction-to-ocaml-ppx-ecosystem) that brings styled components to ReScript. Built on top of [emotion](https://emotion.sh), it allows you to style apps safely, quickly, and with zero runtime overhead - just as you have always done it. styled-ppx allows you to create **React Components** with type-safe style definitions that don't rely on a different language ([DSL](https://en.wikipedia.org/wiki/Domain-specific_language)) except CSS.
```
davesnx, line 7: `with zero runtime overhead`

That's not entirely correct, since emotion brings some. The goal of styled-ppx is to remove emotion eventually, but hasn't been done.

davesnx (approved):

README.md
psb: Okay. I'll change this back to use the word 'performant'. There is some argument that 'performant' isn't a word; however, I think it is used by programmers so it should not be an issue.

davesnx:
> Only single properties are allowed, e.g., color: blue, and multiple declarations, e.g., :hover { color: blue} or color: blue; background-color: red, are not allowed. If you need to use a selector or use more than one rule, use %cx{:rescript} instead.



That make sense

PR #409, opened 2024-01-23
Add `styled-ppx-native` package

jchavarri:
A couple of people in the Melange community have reported some issues while installing `styled-ppx` after the addition of the OCaml native side for universal components (mostly `server-reason-react`).

I think it's nice to keep `styled-ppx` "free" of any native OCaml deps like `server-reason-react`, `pcre` etc, for those users that only need to run it with Melange in browser environments.

This PR adds a new package `styled-ppx-native` that only includes the native parts of it, and removes the `server-reason-react` dep from `styled-ppx` package.


lubegasimon:
> A couple of people in the Melange community have reported some issues while installing styled-ppx after the addition of the OCaml native side for universal components (mostly server-reason-react).

LGTM, for example, what issues?

jchavarri:
> what issues?

mostly problems when installing `styled-ppx` on macos, due to the dependency on `pcre` (which also requires `pkg-config`).

PR #421, opened 2024-02-01
support melange 3 from opam

jchavarri:
Updates `styled-ppx` to use `melange.3.0.0` from opam.

- Update opam file (bump lower bound for melange)
- Remove pin for melange
- Replaces usages of `js` strings in patterns (which triggered an error)
- `deriving abstract` -> `deriving jsProperties, getSet`
- `bs.module` -> `mel.module`

### DIFF
--- a/packages/css/native/shared/Css_AtomicTypes.ml
+++ b/packages/css/native/shared/Css_AtomicTypes.ml
@@ -2277,9 +2277,9 @@ module FontFamilyName = struct
   let toString x =
     match x with
     | `custom value ->
-      (match Std.String.get value 0 with
-      | {js|"|js} | {js|'|js} -> value
-      | _ -> ({js|"|js} ^ value) ^ {js|"|js})
+      let s = Std.String.get value 0 in
+      if s = {js|"|js} || s = {js|'|js} then value
+      else ({js|"|js} ^ value) ^ {js|"|js}
     | `serif -> {js|serif|js}
     | `sansSerif -> {js|sans-serif|js}
     | `cursive -> {js|cursive|js}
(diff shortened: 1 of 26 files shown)


packages/css/native/shared/Css_AtomicTypes.ml
```diff
@@ -2278,7 +2278,7 @@ module FontFamilyName = struct
     match x with
     | `custom value ->
       (match Std.String.get value 0 with
-      | {js|"|js} | {js|'|js} -> value
```
jchavarri, line 2281: Apparently this kind of usage of `js` strings never worked, see [example](https://melange.re/v2.2.0/playground/?language=OCaml&code=let+x+%3D+match+%7Bjs%7C%F0%9F%90%AB+name%7Cjs%7D+with+%7Bjs%7C%F0%9F%90%AB+name%7Cjs%7D+-%3E+%28%29+%7C+_+-%3E+assert+false&live=off).

packages/emotion/js/Css.ml
```diff
@@ -7,12 +7,12 @@ include Css_Legacy_Core.Make (struct
   type renderer = Js.Json.t
 
   external injectRaw : string -> unit = "injectGlobal"
-  [@@mel.module "@emotion/css"] [@@bs.module "@emotion/css"]
+  [@@mel.module "@emotion/css"] [@@mel.module "@emotion/css"]
```
davesnx, line 10: That's breaking ReScript users, I will jump into this PR to make it work since it needs to duplicate this file and it's a little messy right now

jchavarri: Do you think cppo would help here?

davesnx: Oh, also cppo!

davesnx:
Re-opened the same PR from origin https://github.com/davesnx/styled-ppx/pull/423

Issue #357, opened 2023-09-03
Support for media queries level 5


davesnx:
Media queries level 4 are supported, would be nice to support further:

https://www.w3.org/TR/mediaqueries-5/

```css
@media not (width <= -100px) {
  body { background: green; }
}
```

```css
@media (height > -100px) {
  body { background: green; }
}
```

PR #424, opened 2024-02-13
support for media-queries5

lubegasimon:
Support media_types and logical operators in media-queries

Fixes https://github.com/davesnx/styled-ppx/issues/357

### DIFF
--- a/packages/parser/lib/css_lexer.re
+++ b/packages/parser/lib/css_lexer.re
@@ -83,15 +83,33 @@ let number = [%sedlex.regexp?
   (Opt('-'), '.', Plus(digit), Opt('e' | 'E', '+' | '-', Plus(digit)))
 ];
 
-let operator = [%sedlex.regexp? "~=" | "|=" | "^=" | "$=" | "*=" | "="];
+let equal_sign = [%sedlex.regexp? "="];
+let greater_than = [%sedlex.regexp? ">"];
+/* [less_than], [only_operator], [or_operator], [and_operator],
+   [not_operator] are purposely for media query parsing */
+let less_than = [%sedlex.regexp? "<"];
+let only_operator = [%sedlex.regexp? "only"];
+let or_operator = [%sedlex.regexp? "or"];
+let and_operator = [%sedlex.regexp? "and"];
+let not_operator = [%sedlex.regexp? "not"];
 
-let combinator = [%sedlex.regexp? '>' | '+' | '~' | "||"];
+let operator = [%sedlex.regexp? "~=" | "|=" | "^=" | "$=" | "*="];
+
+let combinator = [%sedlex.regexp? '+' | '~' | "||"];
 
 let at_rule_without_body = [%sedlex.regexp?
   ("@", "charset" | "import" | "namespace")
 ];
+
 let at_rule = [%sedlex.regexp? ("@", ident)];
+
+/* [<screen, print, all>_media_type] is purposely for media query parsing */
+let screen_media_type = [%sedlex.regexp? "screen"];
+let print_media_type = [%sedlex.regexp? "print"];
+let all_media_type = [%sedlex.regexp? "all"];
+
 let at_media = [%sedlex.regexp? ("@", "media")];
+
 let at_keyframes = [%sedlex.regexp? ("@", "keyframes")];
 
 let non_ascii_code_point = [%sedlex.regexp? Sub(any, '\000' .. '\128')]; // greater than \u0080
@@ -534,6 +552,16 @@ let rec get_next_token = buf => {
   | combinator => COMBINATOR(latin1(buf))
   | string => STRING(latin1(~skip=1, ~drop=1, buf))
   | important => IMPORTANT
+  | equal_sign => EQUAL_SIGN(latin1(buf))
+  | (or_operator, whitespace)
+  | (and_operator, whitespace)
+  | (not_operator, whitespace)
+  | (only_operator, whitespace) => MEDIA_QUERY_OPERATOR(latin1(buf))
+  | greater_than
+  | less_than => MEDIA_FEATURE_COMPARISON(latin1(buf))
+  | all_media_type => ALL_MEDIA_TYPE(latin1(buf))
+  | screen_media_type => SCREEN_MEDIA_TYPE(latin1(buf))
+  | print_media_type => PRINT_MEDIA_TYPE(latin1(buf))
   | at_media =>
     skip_whitespace.contents = false;
     AT_MEDIA(latin1(~skip=1, buf));
--- a/packages/parser/lib/css_parser.mly
+++ b/packages/parser/lib/css_parser.mly
@@ -32,7 +32,13 @@ open Css_types
 %token <string> NTH_FUNCTION
 %token <string> URL
 %token BAD_URL
+%token <string> EQUAL_SIGN
+%token <string> MEDIA_QUERY_OPERATOR
+%token <string> MEDIA_FEATURE_COMPARISON
 %token <string> AT_MEDIA
+%token <string> SCREEN_MEDIA_TYPE
+%token <string> PRINT_MEDIA_TYPE
+%token <string> ALL_MEDIA_TYPE
 %token <string> AT_KEYFRAMES
 %token <string> AT_RULE
 %token <string> AT_RULE_STATEMENT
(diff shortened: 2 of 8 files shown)


packages/ppx/test/native/At_rule_test.re
```diff
@@ -11,42 +11,36 @@ let media_query_tests = [
...
+    [%expr [%cx "@media (width < 33px) {}"]],
+    [%expr CssJs.style([|CssJs.media({js|(width < 33px)|js}, [||])|])],
+  ),
+  (
+    "(width > 33px)",
+    [%expr [%cx "@media (width > 33px) {}"]],
+    [%expr CssJs.style([|CssJs.media({js|(width > 33px)|js}, [||])|])],
   ),
   (
-    "not all and (monochrome)",
```
davesnx, line 40: Please don't remove test cases, keep them separate

lubegasimon: I have a couple of them locally, I removed those to get a git push through

davesnx: If you want to push broken tests use `git push origin xx --no-verify`

packages/ppx/test/native/At_rule_test.re
```diff
@@ -67,6 +67,13 @@ let media_query_tests = [
       CssJs.style([|CssJs.media({js|screen and (color) |js}, [||])|])
     ],
   ),
+  (
+    "(screen and (color))",
+    [%expr [%cx "@media (screen and (color)) {}"]],
+    [%expr
+      CssJs.style([|CssJs.media({js|(screen and (color))|js}, [||])|])
+    ],
+  ),
```
lubegasimon, line 76: This test fails with `"[%ocaml.error \"Parse error while reading token 'and'\"]`, probably because of the parsing rules, but I am not sure where exactly. Any pointers @davesnx ?

davesnx: Parse error comes from the properties parser, so eventually it would need to understand and as well. There're all media-queries combinations on the properties parser under Parser.re

davesnx: All media_ prefixed from Parser.re are references to media-querie parsing: https://github.com//davesnx/styled-ppx/blob/28035df2f39cd5337f7de5098985f96f769cbf3f/packages/reason-css-parser/lib/Parser.re#L479-L480

packages/parser/lib/css_lexer.re
```diff
@@ -549,9 +549,9 @@ let rec get_next_token = buf => {
   | greater_operator
   | less_operator
   | equal_sign
-  | only_operator // TODO: maybe IDENT
-  // | or_operator // TODO: maybe IDENT
-  | and_operator // TODO: maybe IDENT
+  | only_operator
+  // | or_operator
```
lubegasimon, line 553: This must be a todo; when called, it conflicts with some other `or` which I can't trace atm

packages/parser/lib/css_parser.mly
```diff
@@ -618,3 +618,4 @@ value:
   | v = INTERPOLATION { Variable v } /* $(Lola.value) */
   | f = loc(FUNCTION) v = loc(values) RIGHT_PAREN; { Function (f, v) } /* calc() */
   | u = URL { Uri u } /* url() */
+  | WS { Delim " " }
```
lubegasimon, line 621: This must be unneeded

packages/parser/lib/css_parser.mly
```diff
@@ -89,29 +89,131 @@ bracket_block (X): xs = delimited(LEFT_BRACKET, X, RIGHT_BRACKET) { xs }
...
+/* "and" */
+and_operator: o = OPERATOR { o }
+/* "or" */
+or_operator: o = OPERATOR { o }
+/* "only" */
+only_operator: o = OPERATOR { o }
+/* "not" */
+not: n = OPERATOR { n }
+/* "all" | "screen" | "print" */
+media_type: s = IDENT { s }
```
davesnx, line 109: Maybe have a token for each of the media_types? Feel liks ident would make the parsing more complex here.

packages/parser/lib/css_parser.mly
```diff
@@ -89,29 +89,131 @@ bracket_block (X): xs = delimited(LEFT_BRACKET, X, RIGHT_BRACKET) { xs }
...
+
+/* https://www.w3.org/TR/mediaqueries-5/#mq-syntax */
+
+/* <mf-plain> = <mf-name> : <mf-value> */
```
davesnx, line 113: Would define a certain list of mf-name make the parsing better here?

packages/ppx/test/native/At_rule_test.re
```diff
@@ -5,87 +5,196 @@ let loc = Location.none;
...
       |])
     ],
   ),
-  /* Nested is not supported in code-gen
-     https://www.w3.org/TR/mediaqueries-5/#media-conditions */
-  /* (
-       "(not (color)) and (not (hover))",
-       [%expr [%cx "@media (not (color)) and (not (hover)) {}"]],
-       [%expr CssJs.style( [|CssJs.media( {js|((not (color)) and (not (hover))|js}, [||])|])]
-     ), */
```
davesnx, line 87: Can you keep this test? 

Is there an actual test passing with nesting?

packages/ppx/test/snapshot/reason/media-queries-and-selectors.t/run.t
```diff
@@ -964,7 +964,7 @@
       CssJs.style([|
         CssJs.label("SelectorsMediaQueries"),
         CssJs.media(
-          {js|(min-width: 600px)|js},
+          {js|(min-width: 600px) |js},
```
davesnx, line 967: Why this space is added?

packages/parser/lib/css_lexer.re
```diff
@@ -530,10 +545,18 @@ let rec get_next_token = buf => {
     INTERPOLATION(
       latin1(~skip=2, ~drop=1, buf) |> String.split_on_char('.'),
     )
-  | operator => OPERATOR(latin1(buf))
+  | operator
+  | greater_operator
+  | less_operator
+  | equal_sign
+  | only_operator
```
davesnx, line 552: why are all of those splitted in variables and then lexed into one OPERATOR?

packages/parser/lib/css_parser.mly
```diff
@@ -89,29 +89,131 @@ bracket_block (X): xs = delimited(LEFT_BRACKET, X, RIGHT_BRACKET) { xs }
...
+/* <mf-range> =
+      <mf-name> <mf-comparison> <mf-value>
+      | <mf-value> <mf-comparison> <mf-name>
+      | <mf-value> <mf-lt> <mf-name> <mf-lt> <mf-value>
+      | <mf-value> <mf-gt> <mf-name> <mf-gt> <mf-value> */
+mf_range:
+  | xs = IDENT WS mf_comparison WS value { xs } 
+  | value WS xs =  mf_comparison WS IDENT { xs }
+  | value WS xs = mf_lt WS IDENT WS mf_lt WS value { xs } 
+  | value WS xs = mf_gt WS IDENT WS mf_gt WS value { xs }
```
davesnx, line 137: The values could be interpolations right?

Do you want to add it in a further PR?

packages/parser/lib/css_parser.mly
```diff
@@ -276,7 +380,7 @@ nth_payload:
   /* 2n */
   | a = DIMENSION { Nth (AN (int_of_string (fst a))) }
   /* 2n-1 */
-  | a = DIMENSION WS? combinator = COMBINATOR b = NUMBER {
+  | a = DIMENSION WS? combinator = combinator b = NUMBER {
```
davesnx, line 383: I don't think this is correct. 2n-1 doesn't accept `>`

packages/parser/lib/css_parser.mly
```diff
@@ -300,17 +404,21 @@ nth_payload:
...
+  | n = IDENT WS? combinator = combinator b = NUMBER {
     let first_char = String.get n 0 in
     let a = if first_char = '-' then -1 else 1 in
     Nth (ANB ((a, combinator, int_of_string b)))
   }
   /* TODO: Support "An+B of Selector" */
 
+func:
+  | f = FUNCTION { f }
+  | n = not { n }
```
davesnx, line 416: not isn't a function, why is added like this?

packages/parser/lib/css_parser.mly
```diff
@@ -487,33 +595,6 @@ complex_selector:
...
-  | COMMA { Delim "," }
-  | n = NUMBER { Number n }
-  | r = UNICODE_RANGE { Unicode_range r }
-  | d = FLOAT_DIMENSION { Float_dimension d }
-  | d = DIMENSION { Dimension d }
-  | v = INTERPOLATION { Variable v } /* $(Lola.value) */
-  | f = loc(FUNCTION) xs = loc(prelude) RIGHT_PAREN; { Function (f, xs) } /* calc() */
-  | u = URL { Uri u } /* url() */
-  | WS { Delim " " }
-
```
davesnx, line 516: Does deleting this causes a any problems in terms of conflicts or whitespace stuff?

If no, it's a big improvement

packages/parser/lib/css_parser.mly
davesnx: Yes, that's wrong and I'm afraid you need it since you deleted `value_in_prelude`

packages/ppx/test/native/At_rule_test.re
lubegasimon: Yes, there is one and it passes!

packages/ppx/test/snapshot/reason/media-queries-and-selectors.t/run.t
lubegasimon: it's a bug causing the same in other places too.

packages/parser/lib/css_parser.mly
lubegasimon: > The values could be interpolations right?

Yes, I think I will add it to this PR!

packages/parser/lib/css_parser.mly
lubegasimon: > Does deleting this causes a any problems in terms of conflicts or whitespace stuff?

No, it does not!

lubegasimon:
> Yes, that's wrong and I'm afraid you need it since you deleted value_in_prelude

It's redundant at the moment.

packages/parser/lib/css_parser.mly
lubegasimon: It's redundant at the moment.

packages/parser/lib/css_parser.mly
lubegasimon: That makes sense, I think we need to do the same for the logical operators, `and` & `or`

packages/parser/lib/css_parser.mly
lubegasimon: You're right, thanks!

packages/parser/lib/css_parser.mly
lubegasimon: Absolutely, it would be better if we could somehow call the `Reason` parser's `check_property` somewhere. I'm not sure yet how this will evolve.

packages/parser/lib/css_parser.mly
lubegasimon: right, but we're not parsing it as a sole function in rule, rather as `:not()`; `| COLON f = func xs = paren_block( ... )`

packages/ppx/test/css-support/css-support.expected.re
```diff
@@ -1035,8 +1035,8 @@ CssJs.unsafe({js|colorScheme|js}, {js|light|js});
 CssJs.unsafe({js|colorScheme|js}, {js|dark|js});
 CssJs.unsafe({js|colorScheme|js}, {js|light dark|js});
 CssJs.unsafe({js|colorScheme|js}, {js|dark light|js});
-CssJs.unsafe({js|colorScheme|js}, {js|only light|js});
-CssJs.unsafe({js|colorScheme|js}, {js|light only|js});
```
lubegasimon, line 1039: Not sure whether to remove unsupported properties of the above sort or to implement them in this PR! (the former makes sense to me)

packages/ppx/test/css-support/css-support.expected.re
```diff
@@ -349,7 +349,7 @@ CssJs.cursor(`verticalText);
 CssJs.cursor(`alias);
 CssJs.cursor(`copy);
 CssJs.cursor(`noDrop);
-CssJs.cursor(`notAllowed);
```
lubegasimon, line 352: This can be tweaked to not conflict with `NOT_OPERATOR`, does that make sense?

packages/ppx/test/css-support/css-support.expected.re
davesnx: Does seems problems related on parsing wrongly the property (I believe keeping the component_value_in_prelude is necessary on the parsing face)

packages/ppx/test/css-support/css-support.expected.re
davesnx: It doesn't actually, the not should be parsed in the media prelude, not on the component_value

packages/ppx/test/css-support/css-support.expected.re
```diff
@@ -1594,7 +1594,7 @@ CssJs.unsafe({js|maskType|js}, {js|luminance|js});
 CssJs.unsafe({js|maskType|js}, {js|alpha|js});
 CssJs.unsafe({js|mixBlendMode|js}, {js|normal|js});
 CssJs.unsafe({js|mixBlendMode|js}, {js|multiply|js});
-CssJs.unsafe({js|mixBlendMode|js}, {js|screen|js});
```
davesnx, line 1597: Same

packages/ppx/test/css-support/css-support.expected.re
lubegasimon: Make sense. A hack AFAIU is not to produce `NOT_OPERATOR`, instead we use `IDENT` in the parser which makes the `not` operator unsafe.

davesnx: no, parsing should remain as what It was before (2 separate component_values) and those errors should dissapear

packages/ppx/test/native/At_rule_test.re
```diff
@@ -5,87 +5,216 @@ let loc = Location.none;
...
+  //   ],
+  //   [%expr
+  //     CssJs.style([|
+  //       CssJs.media(
+  //         {js|screen and (min-width: 33px) or (max-height: 15rem)|js},
+  //         [||],
+  //       ),
+  //     |])
+  //   ],
+  // ),
```
lubegasimon, line 196: This too is blocking

packages/ppx/test/native/At_rule_test.re
```diff
@@ -171,15 +171,13 @@ let media_query_tests = [
...
-  //   [%expr [%cx "@media (not (color)) or (hover) {}"]],
-  //   [%expr
-  //     CssJs.style([|
-  //       CssJs.media({js|(not (color)) or (hover)|js}, [||]),
-  //     |])
-  //   ],
-  // ),
+  (
+    "(not (color)) or (hover)",
+    [%expr [%cx "@media (not (color)) or (hover) {}"]],
```
lubegasimon, line 176: Huh! `"@media (not (color)) or(hover) {}` fails

lubegasimon: Surely, this is invalid grammar.

packages/ppx/test/native/At_rule_test.re
```diff
@@ -5,87 +5,200 @@ let loc = Location.none;
...
+  ),
+  (
+    "(screen and (color))",
+    [%expr [%cx "@media (screen and (color)) {}"]],
+    [%expr
+      CssJs.style([|CssJs.media({js|(screen and (color)) |js}, [||])|])
+    ],
+  ),
+  (
+    "(not (screen and (color))))",
```
davesnx, line 100: That's a weird example:

Usually this is used: `not screen and (color)`

packages/ppx/test/native/At_rule_test.re
```diff
@@ -5,87 +5,200 @@ let loc = Location.none;
...
-       [%expr CssJs.style( [|CssJs.media( {js|((not (color)) and (not (hover))|js}, [||])|])]
-     ), */
-  /* "@media screen, print and (color) {}" */
+  (
+    "$(wat)",
+    [%expr [%cx "@media $(wat) {}"]],
+    [%expr CssJs.style([|CssJs.media(wat, [||])|])],
+  ),
+  // (
+  //   "($(wat))",
```
davesnx, line 196: Can you remove this comment?

packages/parser/lib/css_lexer.re
```diff
@@ -534,6 +552,16 @@ let rec get_next_token = buf => {
...
+  | not_operator => NOT_OPERATOR(latin1(buf))
+  | (or_operator, whitespace) => OR_OPERATOR(latin1(buf))
+  | and_operator => AND_OPERATOR(latin1(buf))
+  | only_operator => ONLY_OPERATOR(latin1(buf))
+  | equal_sign => EQUAL_SIGN(latin1(buf))
+  | greater_than => GREATER_THAN(latin1(buf))
+  | less_than => LESS_THAN(latin1(buf))
+  | all_media_type => ALL_MEDIA_TYPE(latin1(buf))
+  | screen_media_type => SCREEN_MEDIA_TYPE(latin1(buf))
+  | print_media_type => PRINT_MEDIA_TYPE(latin1(buf))
```
davesnx, line 564: I'm not a fan of having all operators as different tokens, did you thought about unifying them?

davesnx (approved):
Generally looks good. I would love to see a few lexer tests with component_value and media types.

The only concerning part is the misterious whitespace, that needs to get fixed.

packages/ppx/test/native/At_rule_test.re
```diff
@@ -5,87 +5,200 @@ let loc = Location.none;
...
+  //   [%expr [%cx "@media ($(wat)) {}"]],
+  //   [%expr
+  //     CssJs.style([|CssJs.media({js|(|js} ++ wat ++ {js|)|js}, [||])|])
+  //   ],
+  // ),
 ];
```
davesnx, line 202: We need tests for broken media queries such as:

```
@media (min-width: calc(50vw + 10px)); /* functions aren't valid values */
@media not((color)); /* whitespace must follow `not` */
@media (768px < = width < 1200px); /* cannot have a space between `<` and `=` */
```

packages/parser/lib/css_lexer.re
lubegasimon: Based on https://github.com/davesnx/styled-ppx/pull/424#discussion_r1495995969 to individually generate tokens, but _some_ are unifiable

packages/ppx/test/native/At_rule_test.re
lubegasimon: Thanks, I will add a better example, but also keep the other

packages/ppx/test/native/At_rule_test.re
lubegasimon: > @media (min-width: calc(50vw + 10px)); /* functions aren't valid values */

Why should functions be invalid values?

> @media not((color)); /* whitespace must follow `not` */

This has now been fixed locally, so I expect it to error when written

> @media (768px < = width < 1200px); /* cannot have a space between `<` and `=` */

This currently errors when written

packages/ppx/test/native/At_rule_test.re
lubegasimon: BTW, do we need to keep the above lines or weed them out in this PR?

packages/ppx/test/native/At_rule_test.re
davesnx: functions aren't part of the media-feature, defined by the spec: https://www.w3.org/TR/mediaqueries-5/#mq-features

- Can you add a case where there's no space after not?
- Can you add a case with broken `< =`?

packages/parser/lib/css_lexer.re
```diff
@@ -552,13 +552,13 @@ let rec get_next_token = buf => {
...
   | string => STRING(latin1(~skip=1, ~drop=1, buf))
   | important => IMPORTANT
-  | not_operator => NOT_OPERATOR(latin1(buf))
-  | (or_operator, whitespace) => OR_OPERATOR(latin1(buf))
-  | and_operator => AND_OPERATOR(latin1(buf))
-  | only_operator => ONLY_OPERATOR(latin1(buf))
   | equal_sign => EQUAL_SIGN(latin1(buf))
-  | greater_than => GREATER_THAN(latin1(buf))
-  | less_than => LESS_THAN(latin1(buf))
+  | not_operator => NOT_OPERATOR(latin1(buf))
```
davesnx, line 556: Would it make more sense to have a not_operator always come with a WS after?

This way we can remove the func: from the parser.

lubegasimon: `func` rule is no more

packages/ppx/test/native/At_rule_test.re
lubegasimon: > @media (min-width: calc(50vw + 10px)); /* functions aren't valid values */

That shall be added later because `mf-value` is not parsed safely, so it accepts functions.

davesnx: Can u add a comment on the parser?

lubegasimon: There is a comment on line 166 in the css_parser file

Issue #296, opened 2022-08-04
Allow a prop `as` to change the tag at runtime


davesnx:
Changes the element tag from the React.createElement
We would need to add `as` in makeProps as well

PR #436, opened 2024-03-09
Allow a prop as to change the tag at runtime

zakybilfagih:
My attempt to add `as` props. #296 

For ReScript, I thought of using \\"as" instead of as_, maybe we can handle this conditionally?


packages/ppx/src/generate_lib.re
```diff
@@ -136,32 +136,15 @@ let applyIgnore = (~loc, expr) => {
...
-                    ~loc,
-                    Pconst_string(htmlTag, loc, None),
-                  ),
-                  [],
-                ),
-              ),
-            ),
-          Pconst_string(htmlTag, loc, None),
-        ),
+        Helper.Exp.ident(~loc, withLoc(Lident("finalHtmlTag"), ~loc)),
```
davesnx, line 147: Maybe we can inline the switch expression here, that way we don't rely on anything outside of the createVariadicElement

davesnx (approved):
This looks really good.

In order to have it to work with both ReScript and Reason, the best is to use the same mechanism as https://github.com/davesnx/styled-ppx/blob/main/packages/ppx/src/platform_attributes.re, so in ReScript can keep it as `"as"` I believe.

Left a minor comment about keeping code generated smaller

zakybilfagih:
> In order to have it to work with both ReScript and Reason, the best is to use the same mechanism as https://github.com/davesnx/styled-ppx/blob/main/packages/ppx/src/platform_attributes.re, so in ReScript can keep it as `"as"` I believe.

I can't seem to make `\"as"` work when generating for ReScript ([using illegal identifier names](https://rescript-lang.org/docs/manual/latest/use-illegal-identifier-names)). Since the output should be valid OCaml syntax (?)

Using as_ works fine though.

zakybilfagih:
Managed to get it ot work :)
But some of PPX tests for ReScript is broken. Since `bsc` is outputting invalid OCaml syntax (?)

davesnx:
I have found those issues but they are flaky, I added a few "fixes" like this one https://github.com/davesnx/styled-ppx/pull/436/files#diff-d8703a331a10e5671566f821101c2934d0de98e4464942d485fdd5f03db425b2R4 but still they appear from time to time. Try to `cat` into the res file to see if it's really a syntax issue (and mitigate it or keep the test as it is now)

Once this is fixed, I will merge this, looks great.

zakybilfagih:
I cat'ed the outputted ml file from `bsc`, and confirmed that the error is caused by using `as` as a record field and doing record access (`props.as`).

We can maybe replace those with sed, or is it too hacky?

davesnx:
Right, in ReScript we use record field access for props (`prop.bla`) but in Reason we use `blaGet(prop)`. I wonder if we need to apply a small patch for "as" in the code gen and do: https://rescript-lang.org/try?version=v11.0.1&code=ALBWGcA8GEHsDsBmBLA5gCgN4DcCmAncZBALgAIAWAGjIFtYATXcgIgEMBXAF1lra+QBjFgF8AlACgJXAJ4AHXGXy5BsfAwD6Ad2RcAFhrbgyAXjKYAOu3Aty4LvmTxUIqQBtcXMrkhtacj3JlVXVtXQMjU3MrI1syawZEUSkAKXAAOjdYDB8-ANx0mJtJIA

zakybilfagih:
I've tested the e2e for both v9 and v10 ReScript and the PPX is working correctly (using `as`).

The problem is I think with the ppx cram test.
1. Call `bsc` with the ppx and `input.res` file -> this outputs .ml file
2. Since the AST contains a record field `as` (which the ReScript syntax happily supports) it spits out a record field `as` in OCaml syntax which is not supported
3. When trying to convert back the .ml file using `rescript convert`, it errors out because record field `as` is not valid syntax

davesnx:
I see, then adding a sed (and a comment explaining this) is really good then

davesnx:
Fixed in `main`. Rebase/merge fixes the issue. I could repro locally, can you make sure you have melange.3.0.0 installed?

zakybilfagih:
Ooh yeah I forgot to sync with upstream. Removed the pin and upgraded the deps!

zakybilfagih:
Fixed the cram tests!

Issue #295, opened 2022-08-04
Render more than one global selector on %styled.global


davesnx:
Only one style_rule is allowed in %styled.global: https://styled-ppx.vercel.app/reference/globals

```
%styled.global(`
  html, body {
    margin: 0;
    padding: 0;
  }
`)
```

I would be useful to have support for an entire stylesheet (list of rules)

PR #437, opened 2024-03-11
Render more than one global selector on %styled.global

zakybilfagih:
This should solve #295


packages/ppx/src/css_to_emotion.re
```diff
@@ -374,7 +374,7 @@ and render_selectors = selectors => {
   |> List.map(((selector, _loc)) => render_selector(selector))
   |> String.concat(", ");
 }
-and render_style_rule = (ident, rule: style_rule) => {
+and render_style_rule_parts = (rule: style_rule) => {
```
lubegasimon, line 377: How about if this is defined in `render_style_rule` function and perhaps re-named?

lubegasimon (approved):

packages/ppx/src/css_to_emotion.re
zakybilfagih: I also was thinking to define it inside `render_style_rule` but it is also used on `render_style_rules` 🤔
hehe need to find a better name xD

zakybilfagih: hehe yeah, need to find a better name xD

lubegasimon: > I also was thinking to define it inside render_style_rule but it is also used on render_style_rules 🤔

Right, I didn't notice that.

zakybilfagih: yup definitely better

packages/ppx/src/css_to_emotion.re
```diff
@@ -488,23 +509,13 @@ let render_keyframes = (declarations: rule_list) => {
...
-          %styled.global(" ... ")
-      |},
-    )
-  /* TODO: Add rule to string to finish this error message */
-  };
+  ruleList
+  |> List.filter_map(rule => {
+       switch (rule) {
+       | Style_rule(rule) => Some(rule)
+       | _ => None
```
davesnx, line 516: Maybe we could render at_rules as well?

davesnx (approved):
Looks nice

zakybilfagih:
Refactored the code and also should render at_style now

packages/ppx/src/css_to_emotion.re
```diff
@@ -509,13 +483,18 @@ let render_keyframes = (declarations: rule_list) => {
...
-     })
-  |> render_style_rules(~loc, CssJs.global(~loc))
+  let styles =
+    ruleList
+    |> List.map(rule => {
+         switch (rule) {
+         | Style_rule(style_rule) => render_style_rule(style_rule)
+         | At_rule(at_rule) => render_at_rule(at_rule)
+         | _ =>
+           Generate_lib.error(~loc, "only style rule and at rule supported")
```
davesnx, line 431: Nit: this error only raises when there's a declaration. Can you add an error message like: 

"declarations do not make sense in global, only style rules or @media/@print/etc are supported. If you want to apply styles to everything, try `*` {}"

(^ This is just a placeholder error message)

davesnx (approved):

packages/css/js/Css_Js_Core.ml
```diff
@@ -31,7 +31,7 @@ module type MakeResult = sig
 
   val insertRule : string -> unit
   val renderRule : renderer -> string -> unit
-  val global : string -> rule array -> unit
+  val global : rule array -> unit
```
lubegasimon, line 34: Minor: How about if we use `list` instead?

packages/css/js/Css_Legacy_Core.ml
```diff
@@ -58,7 +58,9 @@ module Make (CssImpl : Css_Core.CssImplementationIntf) :
      and type renderer := CssImpl.renderer = struct
   let insertRule css = CssImpl.injectRaw css
   let renderRule renderer css = CssImpl.renderRaw renderer css
-  let global selector rules = CssImpl.injectRules selector (toJson rules)
+
+  let global rules =
+    CssImpl.injectRules (toJson rules)
```
lubegasimon, line 63: This and I guess some other cases need `dune fmt`ing

packages/ppx/src/css_to_emotion.re
```diff
@@ -426,23 +421,24 @@ let render_keyframes = (declarations: rule_list) => {
...
-  | [Style_rule(rule)] =>
-    render_style_rule(CssJs.global(~loc), rule)
-    |> Generate_lib.applyIgnore(~loc)
-  /* More than one isn't supported by bs-css */
-  | _res =>
-    Generate_lib.error(
-      ~loc,
-      {|
-        `styled.global` only supports one style definition. Transform each definition into a separate styled.global call
+  let onlyStyleRulesAndAtRulesSupported = {|Declarations does not make sense in global styles. Global should consists of style rules or at-rules (e.g @media, @print, etc.)
```
lubegasimon, line 424: How about `Declarations_unsupported`?

lubegasimon (approved):

packages/css/js/Css_Js_Core.ml
davesnx: in CssJs we use arrays everywhere (and moving forward we will get rid of Css which uses lists).

Why do you suggest list here?

lubegasimon: No other reason apart from the imperative nature of arrays over lists.

zakybilfagih: I think this follows the `rule` variant `S`. In Css_Js_Core, `S` is defined `string * rule array` and on Css_Legacy_Core it's `string * rule list`

packages/ppx/src/css_to_emotion.re
zakybilfagih: oo yeah, should be snake case

Issue #441, opened 2024-03-12
Comment on stylesheet not parsed correctly


zakybilfagih:
![comment on stylesheet](https://github.com/davesnx/styled-ppx/assets/55378008/c189bb0d-2e06-453b-906c-3771959c37c9)

davesnx:
Can you open a PR with the tests failing?

zakybilfagih:
Sure!

PR #442, opened 2024-03-12
Stylesheet comments parsed incorrectly

zakybilfagih:
#441

### DIFF
--- a/e2e/melange/src/ui/ui.re
+++ b/e2e/melange/src/ui/ui.re
@@ -1,22 +1,10 @@
 [%styled.global {|
-/* comment */
-/** comment
- multiline 
-*/
   div {
-/** comment
- multiline 
-*/
     background-color: green; /* another one */
   }
 
   @media (min-width: 400px) {
-    /* comment 1 */
     div {
-      padding: 10px;
-/** comment
- multiline 
-*/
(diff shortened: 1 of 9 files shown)


zakybilfagih:
Setting skip_whitespace on parse_stylesheet seems to fix the issue

davesnx:
That's a nice fix, by default we were parsing only one style rule, so I asumed skipping whitespaces was needed. No it isn't anymore 👍🏼

e2e/melange/src/ui/ui.re
```diff
@@ -1,10 +1,22 @@
...
+    background-color: green; /* another one */
   }
 
   @media (min-width: 400px) {
+    /* comment 1 */
     div {
+      padding: 10px;
+/** comment
+ multiline 
+*/
```
davesnx, line 19: Would you mind adding this as a cram test?

davesnx (approved):

e2e/melange/src/ui/ui.re
zakybilfagih: sure!

zakybilfagih:
Found another problem, seems like a bug with `skip_whitespace`
tez123

Issue #432, opened 2024-03-07
Unify CSS parser and lexer


davesnx:
css_parser.mly parses stylesheets/selectors/declarations/properties and values

Parser.re currently parses (and type-checks) properties + values (there's code in Parser.re that might make it possible to parse stylesheets/selectors and everything but isn't used)

Parser.re uses `value.rec` ppx whichs generates the parser fns for each property, while css_parser.mly uses menhir. I'm more inclined on having a hand-made parser or push for menhir Incremental API in order to call the value.rec's parsers, but it's up to see if it's a good idea.

The driver calls one and the other depending at which stage of the parsing happens, in order to have success with https://github.com/davesnx/styled-ppx/issues/429 it's necessary to have more control over the parsing phase.

## Resources

Mix Lexer and Parser like https://github.com/mnxn/eon and take a look at Sedlexing from here: https://github.com/FStarLang/FStar/pull/2203/files

davesnx:
Currently we use 2 lexers that live inside Css_lexer, and run separate test suites: Tokenizer_test and css_lexer_test. 

We need to unify those, expose a single API to tokenize a string of CSS and treat errors as an exception (Css_lexer.Error) (since menhir needs it), run a single test suite.

davesnx:
After the unification of lexers, we have only one. We still have 2 separate parsers with 2 separate techniques (menhir and rules with combinators and the ppx).

Would be nice to join them into the same package. There's a few benefits of having them togeter:

- No more source_of_loc
- Reuse locations, errors and machinery

davesnx:
There's a bit of work left to unify the lexers which is the API, we have from_string and tokenize where one gets a recursive token structure, and the other a list of tokens. Both contain locations, it's a matter to only use one of them

PR #444, opened 2024-03-12
Cleanup Css_lexer from #372

davesnx:
There were a few cleanups left after https://github.com/davesnx/styled-ppx/pull/372 to keep working towards https://github.com/davesnx/styled-ppx/issues/432

- Unified lexing sedlexes
- Added locations on `Css_lexer.parse`, moved that inside the lexer.
- Unreachable doesn't failwith. Also, remove assert(false)
- Rename parse to tokenize

### DIFF
--- a/packages/parser/lib/css_lexer.re
+++ b/packages/parser/lib/css_lexer.re
@@ -5,6 +5,8 @@ module Types = Css_types;
 module Parser = Css_parser;
 module Location = Ppxlib.Location;
 
+let (let.ok) = Result.bind;
+
 /** Signals a lexing error at the provided source location. */
 exception LexingError((Lexing.position, string));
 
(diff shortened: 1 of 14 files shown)


packages/parser/lib/css_lexer.re
```diff
@@ -928,3 +933,51 @@ let from_string = string => {
...
+  );
+};
+
+let to_string = tokens =>
+  tokens
+  |> List.map(((t, _, _)) => render_token(t))
+  |> String.concat(" ")
+  |> String.trim;
+
+let to_debug = tokens =>
```
lubegasimon, line 982: I think this should be defined near its call site since it has one (call site) so far.

packages/parser/lib/css_lexer.re
```diff
@@ -928,3 +933,51 @@ let from_string = string => {
...
+  let pos_end = position_to_string(loc_end);
+  Printf.sprintf(
+    "%s %s..%s",
+    Tokens.token_to_debug(token),
+    pos_start,
+    pos_end,
+  );
+};
+
+let to_string = tokens =>
```
lubegasimon, line 976: I think this should be defined in the file it is called in until there is a need to have it here.

lubegasimon (approved):

packages/parser/lib/css_lexer.re
```diff
@@ -928,3 +933,51 @@ let from_string = string => {
...
+
+let position_to_string = pos =>
+  Printf.sprintf(
+    "[%d,%d+%d]",
+    pos.Lexing.pos_lnum,
+    pos.Lexing.pos_bol,
+    pos.Lexing.pos_cnum - pos.Lexing.pos_bol,
+  );
+
+let debug_token = ((token, loc_start, loc_end)) => {
```
lubegasimon, line 964: Instance of https://github.com/davesnx/styled-ppx/pull/444#discussion_r1522164658

packages/parser/lib/css_lexer.re
davesnx: It's in the interface of the lexer

packages/parser/lib/css_lexer.re
```diff
@@ -5,6 +5,8 @@ module Types = Css_types;
 module Parser = Css_parser;
 module Location = Ppxlib.Location;
 
+let (let.ok) = Result.bind;
```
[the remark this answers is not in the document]
davesnx: Because we use it in a few places, it's almost everywhere.

PR #447, opened 2024-03-13
Add test with failure

davesnx:
I found an issue with the parsing when using 3 interpolation with and. Would you mind taking a look? @lubegasimon


lubegasimon:
> I found an issue with the parsing when using 3 interpolation with and. Would you mind taking a look? @lubegasimon

I don't mind, thanks for the repro.

PR #451, opened 2024-03-15
Support relative selector list for :has pseudo class

zakybilfagih:
#435


davesnx (approved):
Super cool to support Selectors Level 4

Good job!

PR #454, opened 2024-03-17
Migrate css-support to cram tests

davesnx:
This PR moves all css-support into separate cram tests, making sure it type-checks against styled-ppx.css_native and styled-ppx.emotion_native.

I fixed the issues in this PR as well and added a dune_describe_pp to see the preprocessed output (and keep same behaviour as css-supported.expected)


lubegasimon:
Thanks for the refactoring and improvements 👍

Issue #448, opened 2024-03-14
Locations got broken


davesnx:
Unsure when it happened, but between 0.52 and 0.54 the location of the generated code is broken in a few cases, the most annoying is when there's a type-miss match on interpolation:

```
let cosas = `underlin;

[%cx
  {|
  width: fit-content;
  grid-template-columns: fit-content(20px) fit-content(10%);

  @media $(Media.tabletUp) and $(Media.largeDesktopDown) {
    grid-template-columns: fit-content(20px) fit-content(10%);
  }

  grid-template-areas: "bp bp" "os ps";
|}
];

[%cx {|
  text-decoration: $(cosas);
|}];
```

```
+  File "input.re", line 1, characters 1-16:
+  Error: This expression has type [> `underlin ]
+         but an expression was expected of type
+           [< `inherit_
+            | `initial
+            | `lineThrough
+            | `none
+            | `overline
+            | `underline
+            | `unset
+            | `var of string
+            | `varDefault of string * string ]
+         The second variant type does not allow tag(s) `underlin
```

The `line 1, characters 1-16` is wrong, we don't se the right line number and should point to the interpolation instead.

```
[%cx {|
  text-decoration: $(cosas);
                   ^^^^^^^^
|}];
```

davesnx:
Previously we kept track of `container_lnum` when we were parsing on the ppx and it was passed to the code-gen.

davesnx:
Copying the comment from https://github.com/davesnx/styled-ppx/pull/456

After this PR the locations don't look exactly I would like it to, but they are close enough:

```
+  File "input.re", lines 10-11, characters 6-36:
+  10 | ......|
+  11 |     text-decoration: $(cosas).
+  Error: This expression has type [> `underlin ]
+         but an expression was expected of type
+           [< `inherit_
+            | `initial
+            | `lineThrough
+            | `none
+            | `overline
+            | `underline
+            | `unset
+            | `var of string
+            | `varDefault of string * string ]
+         The second variant type does not allow tag(s) `underlin
+  [1]
```

The remaining work to make them perfect is to make sure css_lexer and css_parser respect whitespaces and use loc in menhir properly

--- 

Closing this issue since the bug has been fixed

PR #456, opened 2024-03-20
Improvement for locations in both code-gen and error reporting

davesnx:
Fixes partially https://github.com/davesnx/styled-ppx/issues/448

- Made parser wrapped (Styled_ppx_css_parser)
- Fixed subgrid value on grid-template-columns/rows/auto...
- Rename Css_types to Ast
- Adjust locations before rendering and generating the error
- Removed css-support.expect


davesnx:
After this PR the locations don't look exactly I would like it to, but they are close enough:

```
+  File "input.re", lines 10-11, characters 6-36:
+  10 | ......|
+  11 |     text-decoration: $(cosas).
+  Error: This expression has type [> `underlin ]
+         but an expression was expected of type
+           [< `inherit_
+            | `initial
+            | `lineThrough
+            | `none
+            | `overline
+            | `underline
+            | `unset
+            | `var of string
+            | `varDefault of string * string ]
+         The second variant type does not allow tag(s) `underlin
+  [1]
```

The reamaining work to make them perfect is to make sure css_lexer and css_parser respect whitespaces and use loc in menhir properly

PR #461, opened 2024-03-28
Native support for styled.{{tag}}

zakybilfagih:
When trying to use `styled.{{tag}}` on native, got an error `Uninterpreted extension 'raw'`
This is because `server-reason-react.melange_ppx` doesn't recognize the `raw` extension

### DIFF
--- a/e2e/melange/src/native/ui.re
+++ b/e2e/melange/src/native/ui.re
@@ -2,10 +2,11 @@ let stack = [%cx "display: flex; flex-direction: column"];
 let stackGap = gap => [%cx "gap: $(gap)"];
 
 module Cositas = [%styled.div
-  (~lola=CssJs.px(0)) => {|
+  (~lola=CssJs.px(0), ~id) => {|
   display: flex;
   flex-direction: column;
   gap: $(lola);
+  background-color: $(id);
 |}
 ];
 
@@ -47,7 +48,7 @@ let clx = [%cx
 module App = {
   [@react.component]
   let make = () =>
-    <Cositas as_="section" lola={CssJs.px(10)}>
+    <Cositas as_="section" lola={CssJs.px(10)} id=CssJs.red>
       <div className=clx> {React.string("code everywhere!")} </div>
       <div className=selectors> {React.string("Red text")} </div>
     </Cositas>;
--- a/packages/ppx/test/snapshot/reason/reason-dynamic.t/input.re
+++ b/packages/ppx/test/snapshot/reason/reason-dynamic.t/input.re
@@ -1,6 +1,7 @@
 module DynamicComponent = [%styled.div
-  (~var) => {js|
+  (~var, ~id) => {js|
   color: $(var);
   display: flex;
+  background-color: $(id);
 |js}
 ];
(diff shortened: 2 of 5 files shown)


davesnx:
That's actually a nice improvement, since it fixes the mel.raw issue but you will face other issues with styled.tag in native (mel.raw, bindings such as createVariadicElement).

I would propose to add a flag on the ppx (styled-ppx.lib --native) and have a different creation of elements given the flag.

zakybilfagih:
I see.. I guess the first step is to add `server-reason-react.browser_ppx` library to `styled-ppx.lib`, and set it to add the --js flag when --native flag is not set?

zakybilfagih:
> I see.. I guess the first step is to add `server-reason-react.browser_ppx` library to `styled-ppx.lib`, and set it to add the --js flag when --native flag is not set?
aah nvm, this should be handled by the user (?) we just need to emit valid native code right?

zakybilfagih:
I felt the build time is quite slow (tried building `e2e/melange/native/ui.re` and it felt considerably slower than the non native variant with `styled.{{tag}}`). Can't seem to pin point the slow path.

Is using a lot of ppxlib.metaquot have an impact on build times? It shouldn't affect the build time for the user of the ppx right?

davesnx:
That should have any effect, while building ui (maybe we copy the node_modules by accident?)

packages/ppx/test/snapshot/rescript/rescript-static.t/run.t
```diff
@@ -105,7 +105,7 @@
         contextMenu: string [@bs.optional ];
         controls: bool [@bs.optional ];
         coords: string [@bs.optional ];
-        crossorigin: bool [@bs.optional ];
+        crossOrigin: string [@bs.optional ];
```
davesnx, line 108: Crazy good fix

packages/ppx/test/snapshot/rescript/rescript-static.t/run.t
```diff
@@ -235,7 +235,7 @@
         maxLength: int [@bs.optional ];
         media: string [@bs.optional ];
         mediaGroup: string [@bs.optional ];
-        min: int [@bs.optional ];
+        min: string [@bs.optional ];
```
davesnx, line 238: Same

davesnx (approved):
Looks amazing. Well done

packages/ppx/test/snapshot/rescript/rescript-static.t/run.t
zakybilfagih: Calling `React.domProps` caught this 😆

zakybilfagih:
> That should have any effect, while building ui (maybe we copy the node_modules by accident?)

That doesn't seem to be the case, I tried timing the build command. Just adding the styled.{{tag}} call, cause 2x the build
watzz
firz

davesnx:
Maybe the amount of code generated? Can you debug without any prop?

zakybilfagih:
welp, having a dozen of arguments (ReactDOM.domProps) will inevitably effect the build time. nothing we can do about it

PR #462, opened 2024-04-02
Fix issue 443

### LINKED ISSUE #443: Remove all `bs-css doesn't support` cases from declarations_to_emotion.re

lubegasimon:
Attempts to fix #443

### DIFF
--- a/packages/css/js/Css_Legacy_Core.ml
+++ b/packages/css/js/Css_Legacy_Core.ml
@@ -575,6 +575,7 @@ let fontSize x =
   D
     ( {js|fontSize|js},
       match x with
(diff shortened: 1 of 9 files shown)


packages/css/js/Css_Legacy_Core.ml
```diff
@@ -1726,58 +1726,69 @@ let string_of_borderstyle x =
   | #Var.t as va -> Var.toString va
   | #Cascading.t as c -> Cascading.toString c
```
lubegasimon, line 1728: Not the perfect implementation, but good enough (to me)

PR #467, opened 2024-04-21
Style HTML tag

pedrobslisboa:
## Why

As mentioned by @purefunctor it would be nice to have full SSR support on styled-ppx:
image

That way the user doesn't need to create a function to handle it and it will work out of the box by the library.

## How

- Deprecating `render_style_tag `since it doesn't render a tag but the style rules;
  - This avoids breaking change needs.
- Creating `get_string_style_rules ` based on `render_style_tag`;
- Creating `get_string_style_hashes` to get all hashes;
- Creating the `render_style_html_tag` to deliver a full SSR-supported style tag.
  - Example: `<style data-s data-emotion="css className1">.css-className1 { display: block; }</style>`

### Screenshot

The screenshot below shows how the emotion handles it well, only creating on client size only client styles the universal styles were not recreated.

Captura de Tela 2024-04-25 às 14 54 14

### DIFF
--- a/packages/emotion/native/Css.ml
+++ b/packages/emotion/native/Css.ml
@@ -209,16 +209,16 @@ let resolve_selectors rules =
   in
   rules |> List.map resolve_selector |> List.flatten
 
-let pp_keyframes hash keyframes =
+let pp_keyframes animationName keyframes =
   let pp_keyframe (percentage, rules) =
     Printf.sprintf "%i%% { %s }" percentage (render_declarations rules)
   in
   let definition = keyframes |> List.map pp_keyframe |> String.concat " " in
-  Printf.sprintf "@keyframes %s { %s }" hash definition
+  Printf.sprintf "@keyframes %s { %s }" animationName definition
 
 (* `resolved_rule` here means to print valid CSS, selectors are nested
    and properties aren't autoprefixed. This function transforms into correct CSS. *)
-let pp_rules hash rules =
+let pp_rules className rules =
   (* TODO: Refactor with partition or partition_map. List.filter_map is error prone.
      Ss might need to respect the order of definition, and this breaks the order *)
   let list_of_rules = rules |> resolve_selectors in
@@ -228,19 +228,25 @@ let pp_rules hash rules =
     |> List.flatten
     |> List.filter_map render_declaration
     |> String.concat " "
-    |> fun all -> Printf.sprintf ".%s { %s }" hash all
+    |> fun all -> Printf.sprintf ".%s { %s }" className all
   in
   let selectors =
     list_of_rules
-    |> List.filter_map (render_selectors hash)
+    |> List.filter_map (render_selectors className)
     |> String.concat " "
   in
   Printf.sprintf "%s %s" declarations selectors
 
 type declarations =
   | Globals of rule list
-  | Classnames of rule list
-  | Keyframes of (int * rule list) list
+  | Classnames of {
+      className : string;
+      styles : rule list;
+    }
+  | Keyframes of {
+      animationName : string;
+      keyframes : (int * rule list) list;
+    }
 
 module Stylesheet = struct
   module Hashes = Set.Make (String)
@@ -299,19 +305,19 @@ let keyframes_to_string keyframes =
   in
   keyframes |> List.map pp_keyframe |> String.concat ""
 
-let render_hash prefix hash styles =
+let render_hash hash styles =
   let is_label = function D ("label", value) -> Some value | _ -> None in
   match List.find_map is_label styles with
-  | None -> Printf.sprintf "%s-%s" prefix hash
-  | Some label -> Printf.sprintf "%s-%s-%s" prefix hash label
+  | None -> Printf.sprintf "%s" hash
+  | Some label -> Printf.sprintf "%s-%s" hash label
 
 let style (styles : rule list) =
   match styles with
   | [] -> ""
   | _ ->
-    let hash = Murmur2.default (rules_to_string styles) in
-    let className = render_hash "css" hash styles in
-    Stylesheet.push instance (className, Classnames styles);
+    let hash = render_hash (Murmur2.default (rules_to_string styles)) styles in
+    let className = Printf.sprintf "%s-%s" "css" hash in
+    Stylesheet.push instance (hash, Classnames { className; styles });
     className
 
 let global (styles : rule list) =
@@ -328,24 +334,40 @@ let keyframes (keyframes : (int * rule list) list) =
   | _ ->
     let hash = Murmur2.default (keyframes_to_string keyframes) in
     let animationName = Printf.sprintf "%s-%s" "animation" hash in
-    Stylesheet.push instance (animationName, Keyframes keyframes);
+    Stylesheet.push instance (hash, Keyframes { animationName; keyframes });
     animationName
 
 (** Deprecated: Use get_style_rules instead*)
 let render_style_tag () =
   Stylesheet.get_all instance
   |> List.fold_left
-       (fun accumulator (hash, rules) ->
+       (fun accumulator (_, rules) ->
          match rules with
          | Globals rules ->
            Printf.sprintf "%s %s" accumulator (rules_to_string rules)
-         | Classnames rules ->
-           let rules = pp_rules hash rules |> String.trim in
+         | Classnames { className; styles } ->
+           let rules = pp_rules className styles |> String.trim in
            Printf.sprintf "%s %s" accumulator rules
-         | Keyframes keyframes ->
-           let rules = pp_keyframes hash keyframes |> String.trim in
+         | Keyframes { animationName; keyframes } ->
+           let rules = pp_keyframes animationName keyframes |> String.trim in
            Printf.sprintf "%s %s" accumulator rules)
        ""
   |> String.trim
 
 let get_string_style_rules = render_style_tag
+
+let get_string_style_hashes () =
+  Stylesheet.get_all instance
+  |> List.fold_left
+       (fun accumulator (hash, _) ->
+         Printf.sprintf "%s %s" accumulator hash |> String.trim)
+       ""
+
+let style_tag ?key:_ ?children:_ () =
+  React.createElement "style"
+    [
+      String ("data-emotion", "css " ^ get_string_style_hashes ());
+      Bool ("data-s", true);
+      DangerouslyInnerHtml (get_string_style_rules ());
+    ]
+    []
--- a/packages/emotion/native/Css.ml
+++ b/packages/emotion/native/Css.ml
@@ -316,12 +316,11 @@ let style (styles : rule list) =
 
 let global (styles : rule list) =
   match styles with
-  | [] -> "";
+  | [] -> ""
   | _ ->
     let hash = Murmur2.default (rules_to_string styles) in
     Stylesheet.push instance (hash, Globals styles);
     hash
-    
 
 let keyframes (keyframes : (int * rule list) list) =
   match keyframes with
@@ -332,6 +331,7 @@ let keyframes (keyframes : (int * rule list) list) =
     Stylesheet.push instance (animationName, Keyframes keyframes);
     animationName
 
+(** Deprecated: Use get_style_rules instead*)
 let render_style_tag () =
   Stylesheet.get_all instance
   |> List.fold_left
@@ -347,3 +347,5 @@ let render_style_tag () =
            Printf.sprintf "%s %s" accumulator rules)
        ""
   |> String.trim
+
+let get_string_style_rules = render_style_tag
--- a/e2e/melange/debug.re
+++ b/e2e/melange/debug.re
@@ -2,9 +2,8 @@ let _ = [%cx {|
   justify-content: center;
 |}];
 
+print_endline("Rendered app:");
 print_endline(Ui_native.Ui.getStaticMarkup());
 
-print_endline("");
-print_endline("<style>");
-print_endline(CssJs.render_style_tag());
-print_endline("</style>");
+print_endline("\nStyle tag:");
+print_endline(ReactDOM.renderToStaticMarkup(<CssJs.style_tag />));
--- a/e2e/melange/dune
+++ b/e2e/melange/dune
@@ -12,6 +12,6 @@
  (name debug)
  (public_name e2e_melange_debug)
  (modules :standard \ index)
- (libraries ui_native styled-ppx.emotion_native styled-ppx.css_native)
+ (libraries ui_native server-reason-react.react server-reason-react.reactDom styled-ppx.emotion_native styled-ppx.css_native)
  (preprocess
   (pps styled-ppx)))
--- a/e2e/melange/src/native/ui.re
+++ b/e2e/melange/src/native/ui.re
@@ -1,6 +1,19 @@
+[%styled.global
+  {|
+  div {
+    background-color: green;
+  }
+
+  @media (min-width: 400px) {
+    div {
+      background-color: red;
+    }
+  }
+|}
+];
+
 let stack = [%cx "display: flex; flex-direction: column"];
 let stackGap = gap => [%cx "gap: $(gap)"];
-
 module Cositas = [%styled.div
   (~lola=CssJs.px(0), ~id) => {|
   display: flex;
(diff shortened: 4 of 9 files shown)


packages/emotion/test/test_css_js_tag.ml
```diff
@@ -0,0 +1,345 @@
...
+
+let multiple_properties () =
+  let className =
+    CssJs.style [| CssJs.display `block; CssJs.fontSize (`px 10) |]
+  in
+  let css = render_style_html_tag () in
+  assert_string css
+    (Printf.sprintf
+       "<style data-s data-emotion=\"css %s\">.%s { display: block; font-size: \
+        10px; }</style>"
```
davesnx, line 28: We might not need the same tests cases as test_css_js_styles and we can live with one simple case and a "real world" example.

packages/emotion/native/Css.ml
```diff
@@ -335,3 +336,36 @@ let render_style_tag () =
...
+             pp_keyframes hash keyframes
+             |> Printf.sprintf "%s %s" prevStyles
+             |> String.trim
+           in
+           let dataEmotion =
+             Printf.sprintf "%s %s" prevDataEmotion hash |> String.trim
+           in
+           dataEmotion, rules)
+       ("css", "")
+  |> (fun (dataEmotion, styles) ->
```
davesnx, line 368: Can you use render_style_tag inside render_style_html_tag?

packages/emotion/test/test.ml
```diff
@@ -5,4 +5,6 @@ let () =
       Test_css_styles.tests;
       Test_css_autoprefixer.tests;
       Test_css_hash.tests;
+      Test_css_js_tag.tests;
+      Test_css_tag.tests;
```
davesnx, line 9: Having both tests is a bit unnecessary, I would only keep one

davesnx (commented):
Thanks for doing this, it's a great improvemenet

packages/emotion/native/Css.ml
pedrobslisboa: Hey @davesnx , to use the `render_style_tag` we would need something like this:

```ocaml
(** Deprecated: Use get_string_style_rules instead*)
let render_style_tag () =
  Stylesheet.get_all instance
  |> List.fold_left
       (fun accumulator (hash, rules) ->
         match rules with
         | Classnames rules ->
           let rules = pp_rules hash rules |> String.trim in
           Printf.sprintf "%s %s" accumulator rules
         | Keyframes keyframes ->
           let rules = pp_keyframes hash keyframes |> String.trim in
           Printf.sprintf "%s %s" accumulator rules)
       ""
  |> String.trim

let get_string_style_rules = render_style_tag

let get_string_style_hashes () =
  Stylesheet.get_all instance
  |> List.fold_left
       (fun accumulator (hash, _) ->
         Printf.sprintf "%s %s" accumulator hash |> String.trim)
       ""

let render_style_html_tag () =
  Printf.sprintf "<style data-s data-emotion=\"css %s\">%s</style>"
    (get_string_style_hashes ()) (get_string_style_rules ())

```
WDYT?

I'm also trying to define a better name for those functions, any idea about it?

purefunctor:
This looks good so far! Though, there's one quirk about the style tag rendering that I misspoke about in the screenshot above. The expected output should actually look like this:

```html
<style data-emotion="css id123-someName" data-s>
  .css-id123-someName { }
</style>
```

Basically, the IDs used in `data-emotion` should not have the `css-` prefix. Also, to keep consistent with emotion's client-side behaviour maybe `data-s` could go second?

To my knowledge, `Stylesheet.get_all` already returns names prefixed with `css-`, so the only way to extract the IDs is to mechanically chop the prefix, which works, but is a little unsafe. Thoughts, @davesnx?

davesnx:
Probably best to add the prefix where it matters only, and render here without any change.

Do you know why emotion doesn't need the "css-" prefix here?

pedrobslisboa:
> This looks good so far! Though, there's one quirk about the style tag rendering that I misspoke about in the screenshot above. The expected output should actually look like this:
> 
> ```
> <style data-emotion="css id123-someName" data-s>
>   .css-id123-someName { }
> </style>
> ```
> 
> Basically, the IDs used in `data-emotion` should not have the `css-` prefix. Also, to keep consistent with emotion's client-side behaviour maybe `data-s` could go second?
> 
> To my knowledge, `Stylesheet.get_all` already returns names prefixed with `css-`, so the only way to extract the IDs is to mechanically chop the prefix, which works, but is a little unsafe. Thoughts, @davesnx?

@purefunctor What about the `animation-` prefix? Does it need to be removed?

purefunctor:
> Do you know why emotion doesn't need the "css-" prefix here?

I'm not entirely sure myself, maybe it's just how emotion encodes it? Ultimately it sees the prefix already as the first word under `data-emotion`.

> What about the animation- prefix? Does it need to be removed?

Not sure as well, it's be good to check but I won't be able to until tomorrow!

pedrobslisboa:
> Probably best to add the prefix where it matters only, and render here without any change.
> 
> Do you know why emotion doesn't need the "css-" prefix here?

Were you talking about something like this?

```diff
- let pp_keyframes hash keyframes =
+ let pp_keyframes preffix hash keyframes =
   let pp_keyframe (percentage, rules) =
     Printf.sprintf "%i%% { %s }" percentage (render_declarations rules)
   in
   let definition = keyframes |> List.map pp_keyframe |> String.concat " " in
-  Printf.sprintf "@keyframes %s { %s }" hash definition
+  Printf.sprintf "@keyframes %s-%s { %s }" preffix hash definition

(* ... *)

(* `resolved_rule` here means to print valid CSS, selectors are nested
   and properties aren't autoprefixed. This function transforms into correct CSS. *)
- let pp_rules hash rules =
+ let pp_keyframes preffix hash keyframes =
   (* TODO: Refactor with partition or partition_map. List.filter_map is error prone.
      Ss might need to respect the order of definition, and this breaks the order *)
   let list_of_rules = rules |> resolve_selectors in
   let declarations =
     list_of_rules
     |> List.map Autoprefixer.prefix
     |> List.flatten
     |> List.filter_map render_declaration
     |> String.concat " "
-    |> fun all -> Printf.sprintf ".%s { %s }" hash all
+    |> Printf.sprintf "@keyframes %s-%s { %s }" preffix hash definition
   in

(* ... *)

- let render_hash prefix hash styles =
+ let render_hash hash styles =
    let is_label = function D ("label", value) -> Some value | _ -> None in
    match List.find_map is_label styles with
-   | None -> Printf.sprintf "%s-%s" prefix hash
-   | Some label -> Printf.sprintf "%s-%s-%s" prefix hash label
+   | None -> Printf.sprintf "%s" hash
+   | Some label -> Printf.sprintf "%s-%s" hash label

let style (styles : rule list) =
  match styles with
  | [] -> ""
  | _ ->
+    let type_ = "css"
     let hash = Emotion_hash.Hash.default (rules_to_string styles) in
-    let className = render_hash "css" hash styles in
+    let className = render_hash hash styles in
     Stylesheet.push instance (type_, className, Classnames styles);
-    className
+    Printf.sprintf "%s-%s" type_ className

let keyframes (keyframes : (int * rule list) list) =
  match keyframes with
  | [] -> ""
  | _ ->
+  let type_ = "animation"
    let hash = Emotion_hash.Hash.default (keyframes_to_string keyframes) in
    Stylesheet.push instance (hash, Keyframes keyframes);
-   animationName
+  "Printf.sprintf "%s-%s" type_ animationName

let get_style_rules () =
  Stylesheet.get_all instance
  |> List.fold_left
        (fun accumulator (type_, hash, rules) ->
          match rules with
          | Classnames rules ->
            let rules = pp_rules type_ hash rules |> String.trim in
            Printf.sprintf "%s %s" hash rules
          | Keyframes keyframes ->
            let rules = pp_keyframes type_ hash keyframes |> String.trim in
            Printf.sprintf "%s %s" hash rules)
        ""
  |> String.trim
```

purefunctor:
Here's an initial fix I came up with, it doesn't account for animations yet:
https://github.com/purefunctor/styled-ppx/commit/3032deec06c52be6ab2f6997ddeeda1d40d3f5fd

purefunctor:
Okay, so with animations, here's what `@emotion/server` emits on my SSR script. They're stuffed under the `css-global` namespace with the hashes (as expected):

```html
<style data-emotion="css-global 1acbpvw">
  @-webkit-keyframes animation-1acbpvw {
    0% {
      opacity: 0;
    }

    100% {
      opacity: 1;
    }
  }

  @keyframes animation-1acbpvw {
    0% {
      opacity: 0;
    }

    100% {
      opacity: 1;
    }
  }
</style>
```

I also discovered that for `[%styled.global ...]` styles, emotion emits the following, though I'm not entirely sure where it got the hash from.
```html
<style data-emotion="css-global 1dvid4v">
  body {
    background-color: #1F1F1F;
    color: #F5F5F5;
    margin: 0;
    padding: 0;
  }

  #root {
    display: -webkit-box;
    display: -webkit-flex;
    display: -ms-flexbox;
    display: flex;
    -webkit-flex-direction: column;
    -ms-flex-direction: column;
    flex-direction: column;
    box-sizing: border-box;
    padding: 1rem;
    min-height: 100vh;
  }

  #root>main {
    -webkit-box-flex: 1;
    -webkit-flex-grow: 1;
    -ms-flex-positive: 1;
    flex-grow: 1;
  }

  @media screen and (min-width: 768px) {
    #root {
      max-width: 900px;
      margin-left: auto;
      margin-right: auto;
    }
  }
</style>
```

pedrobslisboa:
⚠️ To my future me ⚠️ 

It would be nice to have a demo of this fature on styleppx

packages/emotion/native/Css.ml
```diff
@@ -228,31 +228,33 @@ let pp_rules hash rules =
...
   | Classnames of rule list
   | Keyframes of (int * rule list) list
 
 module Stylesheet = struct
   module Hashes = Set.Make (String)
 
   type 'a t = {
-    mutable rules : (string * declarations) list;
+    (* rules: (prefix * hash * rules) *)
+    mutable rules : (string * string * declarations) list;
```
davesnx, line 250: maybe a record here would be needed now, wdyt?

packages/emotion/native/Css.ml
```diff
@@ -298,40 +300,67 @@ let keyframes_to_string keyframes =
...
+    let className = render_hash hash styles in
+    Stylesheet.push instance (preffix, className, Classnames styles);
+    Printf.sprintf "%s-%s" preffix className
+
+let global (styles : rule list) =
+  match styles with
+  | [] -> ""
+  | _ ->
+    let hash = Murmur2.default (rules_to_string styles) in
+    Stylesheet.push instance ("", hash, Global styles);
```
davesnx, line 324: Looking at this I wonder if the prefix can be stored inside Classnames and don't need to do anything in Global, thoughts?

packages/emotion/native/Css.ml
```diff
@@ -298,40 +300,67 @@ let keyframes_to_string keyframes =
...
+let get_string_style_rules = render_style_tag
+
+let get_string_style_hashes () =
+  Stylesheet.get_all instance
+  |> List.fold_left
+       (fun accumulator (_, hash, _) ->
+         Printf.sprintf "%s %s" accumulator hash |> String.trim)
+       ""
+
+let render_style_html_tag () =
```
davesnx, line 363: You moved away from the idea of having a component here?

pedrobslisboa: I forgot to add it.
I'm going to apply it like this:

```ocaml
let style_tag =
  React.createElement "style"
    [
      String ("data-emotion", "css " ^ get_string_style_hashes ());
      Bool ("data-s", true);
      DangerouslyInnerHtml (get_string_style_rules ());
    ]
    []
```

Works for you?

packages/emotion/native/Css.ml
pedrobslisboa: Nice idea, what about something like this:

```ocaml
let style (styles : rule list) =
  match styles with
  | [] -> ""
  | _ ->
    let hash = render_hash (Murmur2.default (rules_to_string styles)) styles in
    let className = Printf.sprintf "%s-%s" "css" hash in
    Stylesheet.push instance (hash, Classnames (className, styles));
    className

let global (styles : rule list) =
  match styles with
  | [] -> ""
  | _ ->
    let hash = Murmur2.default (rules_to_string styles) in
    Stylesheet.push instance (hash, Global styles);
    hash

let keyframes (keyframes : (int * rule list) list) =
  match keyframes with
  | [] -> ""
  | _ ->
    let hash = Murmur2.default (keyframes_to_string keyframes) in
    let animationName = Printf.sprintf "%s-%s" "animation" hash in
    Stylesheet.push instance (hash, Keyframes (animationName, keyframes));
    animationName
```

Then we can manage hash, which is a value for all styles, and the variant provides the need for extra info.

The usage can be:

```ocaml
let render_style_tag () =
  Stylesheet.get_all instance
  |> List.fold_left
       (fun accumulator (_, rules) ->
         match rules with
         | Global rules ->
           Printf.sprintf "%s %s" accumulator (rules_to_string rules)
         | Classnames (className, rules) ->
           let rules = pp_rules className rules |> String.trim in
           Printf.sprintf "%s %s" accumulator rules
         | Keyframes (animationName, keyframes) ->
           let rules = pp_keyframes animationName keyframes |> String.trim in
           Printf.sprintf "%s %s" accumulator rules)
       ""
  |> String.trim
```

packages/emotion/native/Css.ml
pedrobslisboa: @davesnx I was trying to create this style_tag to use like <Css.styleTag/>

```ocaml
let style_tag ?key:_ ()=
  React.createElement "style"
    [
      String ("data-emotion", "css " ^ get_string_style_hashes ());
      Bool ("data-s", true);
      DangerouslyInnerHtml (get_string_style_rules ());
    ]
    []
```

But I'm receiving this issue:

image

A workaround for it is:

```ocaml
let style_tag ?key:_ ~children:_ ()=
  React.createElement "style"
    [
      String ("data-emotion", "css " ^ get_string_style_hashes ());
      Bool ("data-s", true);
      DangerouslyInnerHtml (get_string_style_rules ());
    ]
    []
```

But seems strange to have children that isn't used and even without breaking the calling when adding it:
`<Css.styleTag/> children </Css.styleTag>` It should have an error since it shouldn't receive a children value.

Should I use the workaround or do you have any idea how to handle it?

packages/emotion/native/Css.ml
pedrobslisboa: Due to your comment below (https://github.com/davesnx/styled-ppx/pull/467#discussion_r1580874013)

I was think to keep as before:
```suggestion
    mutable rules : (string * declarations) list;
```

And have the declarations like this:
```ocaml
type declarations =
  | Global of rule list
  | Classnames of (string * rule list)
  | Keyframes of (string * (int * rule list) list)
```

Maybe declarations Classnames and Keyframes could be a record for better documentation.

```ocaml
type declarations =
  | Global of rule list
  | Classnames of {
      className : string;
      styles : rule list;
    }
  | Keyframes of {
      animationName : string;
      keyframes : (int * rule list) list;
    }

```

WDYT?

packages/emotion/native/Css.ml
davesnx: The error message isn't the best, but it tells about a missing `children` argument, which should be there, so the function should be:

```
let style_tag ?key:_ () =
  React.createElement "style"
    [
      String ("data-emotion", "css " ^ get_string_style_hashes ());
      Bool ("data-s", true);
      DangerouslyInnerHtml (get_string_style_rules ());
    ]
    []
```

actually, not even key should be needed in this case, it doesn't hurt but isn't needed by any chance.

pedrobslisboa: I thought the same, but I guess it is the way JSX is being handled, when calling the make function from the component as <Component />, it expects those parameters. Does it make sense?

davesnx: Random idea: Do you have server-reason-react.react added as dependency?

pedrobslisboa: I'm testing it on e2e-melange which has this dune for debug:
```
(executable
 (name debug)
 (public_name e2e_melange_debug)
 (modules :standard \ index)
 (libraries ui_native server-reason-react.react server-reason-react.reactDom styled-ppx.emotion_native styled-ppx.css_native)
 (preprocess
  (pps styled-ppx)))
```

davesnx (approved):
Really great addition.

PR #468, opened 2024-04-25
Global styles for native server on emotion

pedrobslisboa:
## Why

The global styling wasn't available on the server side.

## How

- Create a `global` function to handle global styles on the server side;
- The `global` provides its hash to a future SSR style tag delivery.

### DIFF
--- a/packages/emotion/native/Css.ml
+++ b/packages/emotion/native/Css.ml
@@ -238,6 +238,7 @@ let pp_rules hash rules =
   Printf.sprintf "%s %s" declarations selectors
 
 type declarations =
+  | Globals of rule list
   | Classnames of rule list
   | Keyframes of (int * rule list) list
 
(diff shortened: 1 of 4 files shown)


davesnx (approved):
That's a super good addition, the fact that globals are handled different was something I didn't expect.

Nice work

packages/emotion/native/Css.ml
```diff
@@ -238,6 +238,7 @@ let pp_rules hash rules =
   Printf.sprintf "%s %s" declarations selectors
 
 type declarations =
+  | Global of rule list
```
pedrobslisboa, line 241: @davesnx I know that you approved it, but I have a request change to myself 😆 

Maybe Globals is a more precisely name for it:
```suggestion
  | Globals of rule list
```

davesnx: Sounds good

pedrobslisboa: @davesnx Done ✅

PR #471, opened 2024-05-28
Transform with variable handle unsafe interpolation

### LINKED ISSUE #412: Interpolation renders wrongly when the property isn't supported
This should fail to parse/render and avoid the code-generation:

screenshot_2023-12-11_at_13 21 23

zakybilfagih:
Fixes #412
1. `Combinator.combine_xor` will match the longest rule, so although `Standard.interpolation` matches the value, it will select the parser rule. Added `Combinator.combine_xor_first`, which will select the first successful rule.
2. Pass the interpolation expression using `Unsafe_interpolation` exn so we can render them at `render_when_unsupported_features`

### DIFF
--- a/packages/ppx/src/declarations_to_emotion.re
+++ b/packages/ppx/src/declarations_to_emotion.re
@@ -22,6 +22,7 @@ let (let.ok) = Result.bind;
 /* TODO: Separate unsupported_feature from bs-css doesn't support or can't interpolate on those */
 /* TODO: Add payload on those exceptions */
 exception Unsupported_feature;
+exception Unsafe_interpolation(expression);
 
 exception Invalid_value(string);
 
@@ -75,7 +76,7 @@ let transform_with_variable = (parser, mapper, value_to_expr) =>
   emit(
     /* This Xor is defined here for those properties that aren't defined with
        <interpolation> as a valid definition */
-    Combinator.combine_xor([
+    Combinator.combine_xor_first([
       /* If the entire CSS value is interpolated, we treat it as a `Variable */
       Rule.Match.map(Standard.interpolation, data => `Variable(data)),
       /* Otherwise it's a regular CSS `Value and match the parser */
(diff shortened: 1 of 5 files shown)


packages/ppx/test/native/Interpolation_test.re
```diff
@@ -172,6 +172,17 @@ let tests =
...
     ),
+    (
+      "column-gap: $(Size.px30);",
+      [%expr [%css "column-gap: $(Size.px30)"]],
+      [%expr CssJs.unsafe({js|columnGap|js}, Size.px30)],
+    ),
+    (
+      "padding-inline: $(Size.px30);",
+      [%expr [%css "padding-inline: $(Size.px30)"]],
+      // This is because padding-inline is not inside the properties list on declarations_to_emotion.re
```
davesnx, line 183: Good test, there shouldn't be any case like this in the future

packages/ppx/src/declarations_to_emotion.re
```diff
@@ -3006,7 +3007,7 @@ let transition_delay =
   );
 let transition =
   unsupportedValue(Parser.property_transition, (~loc) =>
-    [%expr CssJs.transition]
+    [%expr CssJs.transitionList]
```
davesnx, line 3010: just to doble check, that's on purpose right?

packages/reason-css-parser/lib/Combinator.re
```diff
@@ -40,6 +53,16 @@ let combine_xor =
       value;
     };
 
+let combine_xor_first =
```
davesnx, line 56: this could be named "first", right?

davesnx (approved):
Nice!

packages/reason-css-parser/lib/Combinator.re
zakybilfagih: hehe yeah

packages/ppx/test/native/Interpolation_test.re
zakybilfagih: yup, should add the missing ones as unsupportedProperty

davesnx: let's do it in another pr!

PR #472, opened 2024-05-28
Add support for transition

zakybilfagih:


packages/ppx/src/declarations_to_emotion.re
```diff
@@ -3004,9 +3013,46 @@ let transition_delay =
...
-  unsupportedValue(Parser.property_transition, (~loc) =>
-    [%expr CssJs.transition]
+  monomorphic(
+    Parser.property_transition,
+    (~loc) => [%expr CssJs.transitionList],
+    (~loc) =>
+      fun
+      | [] => raise(Invalid_value("expected at least one argument"))
+      | [x, ...xs] => {
+          let transitions = [x] @ List.map(Fun.id, xs);
```
davesnx, line 3052: tiny improvement

```suggestion
			| values => 
					values
					|> List.map(render_single_transition(~loc))
          |> Builder.pexp_array(~loc);
```

packages/ppx/test/css-support/transitions.t/run.t
```diff
@@ -55,4 +55,60 @@ This test only runs against Css_Js_Core from styled-ppx.css_native
...
+    ),
+  |]);
+  CssJs.transitionList([|
+    CssJs.Transition.shorthand(
+      ~duration=?Some(`s(1)),
+      ~delay=?Some(`ms(500)),
+      ~timingFunction=?Some(`linear),
+      {js|margin-right|js},
+    ),
+  |]);
```
davesnx, line 114: Would be nice to add a few tests under css-support (packages/ppx/test/css-support) to make sure the interface (CssJs) type-checks and also a few tests cases with interpolation, which might be a bit tricky to get right since we use "||" for the spec.

zakybilfagih: yup deffo needed

zakybilfagih: added!

davesnx (approved):
LFG

PR #473, opened 2024-05-29
Add depext for @emotion/css >= 11.0.0

feihong:
This way, running `opam exec opam-check-npm-deps` will return output for `styled-ppx` and warn you if your version of `@emotion/css` is absent or too old.


davesnx:
Cool

PR #478, opened 2024-06-07
Fix error line number coming from parser

zakybilfagih:


davesnx (approved):

packages/ppx/test/css-support/non-valid-property-location-error-nested.t/run.t
```diff
@@ -0,0 +1,55 @@
...
+
+  $ cat > dune << EOF
+  > (executable
+  >  (name input)
+  >  (libraries styled-ppx.emotion_native styled-ppx.css_native)
+  >  (preprocess (pps styled-ppx)))
+  > EOF
+
+  $ dune build
+  File "input.re", lines 5-6, characters 13-13:
```
davesnx, line 18: are columns correct here?

davesnx (approved):

packages/ppx/test/css-support/non-valid-property-location-error-nested.t/run.t
zakybilfagih: yup all correct now

davesnx: seems like characters 13-13 is wrong thought

zakybilfagih: yeah, this is because the declaration name loc_start is off from the parser. i've fixed it by changing it based on the loc_end and the property string length

PR #479, opened 2024-06-19
Replace usages of `j` and `js` quoted strings with ones without id

jchavarri:
In https://github.com/ml-in-barcelona/server-reason-react/pull/139, server-reason-react is starting to fail if it finds `j` or `js` in quoted strings for code that uses `melange.ppx`.

Using that patch, I noticed some of the code in styled-ppx was using these quoted strings, but there's no unicode or any interpolation, so it can just use quoted strings without id `{||}`.

The diff is long, but the process was mostly find and replace on the affected files, nothing else. I originally started using regular strings, but found a few places where double quotes are used, so I feel safer by keeping the same literals without having to escape.


jchavarri:
Closing this as not necessary anymore, see https://github.com/ml-in-barcelona/server-reason-react/pull/139#issuecomment-2180873831.

PR #480, opened 2024-06-23
Docs: Explain show server rendered stylesheets work natively

ManasJayanth:
We're using styled-ppx at our meetup today and thought we'd contribute some docs. All credits go to @davesnx as he gave a heads up. Thought I'd return the favour ❤️


davesnx (approved):
Thanks, that's a good piece of docs.

PR #483, opened 2024-06-28
Fix nesting for selectors (and pseudo) in native

davesnx:
Resolves all issues with nesting media-queries and selectors when using emotion_native defined here: https://github.com/davesnx/styled-ppx/issues/359

### DIFF
--- a/packages/emotion/native/Css.ml
+++ b/packages/emotion/native/Css.ml
@@ -186,6 +186,7 @@ let resolve_selectors rules =
   (* media selectors should be at the top. .a { @media () {} }
      should be @media () { .a {}} *)
   (* TODO: Only works with 2 levels *)
+  (* TODO: Should push to top any selector *)
   let move_media_at_top rule_list =
     List.fold_left
       (fun acc rule ->
--- a/packages/emotion/native/Css.ml
+++ b/packages/emotion/native/Css.ml
@@ -401,7 +401,7 @@ let flush () = Stylesheet.flush instance
 
 let style (styles : rule list) =
   match styles with
-  | [] -> ""
+  | [] -> "css-0"
   | _ ->
     let hash = render_hash (Murmur2.default (rules_to_string styles)) styles in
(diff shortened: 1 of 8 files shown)


packages/emotion/native/Css.ml
```diff
@@ -107,6 +107,7 @@ let render_declarations (rules : rule list) =
 let is_at_rule selector = String.contains selector '@'
 let is_a_pseudo_selector selector = String.starts_with ~prefix:":" selector
 let starts_with_at selector = String.starts_with ~prefix:"@" selector
+let starts_with_double_dot selector = String.starts_with ~prefix:":" selector
```
pedrobslisboa, line 110: Dumb question: what is the difference between `starts_with_double_dot` and `is_a_pseudo_selector`

pedrobslisboa: Just saw you remove it here: https://github.com/davesnx/styled-ppx/pull/483/files#diff-fae8d1b5ad79dc3019cf34c19b6577b5e210c318ad4fe23e9816de4ad4c0a4daL140

Should we do the same here?

packages/emotion/test/test_css_styles.ml
```diff
@@ -179,50 +179,50 @@ let media_queries_nested_2 =
...
+           ];
+       ]
+   in
+   let css = get_string_style_rules () in
+   assert_string css
+     (Printf.sprintf
+        ".%s { max-width: 800px; } @media (min-width: 300px) { .%s { position: \
+         fixed; }  } @media (max-width: 768px) and (min-width: 300px) { .%s { \
+         display: flex; }  }"
+        classname classname classname) *)
```
pedrobslisboa, line 203: Did something break on this test?

packages/emotion/native/Css.ml
davesnx: I'm removing support for Css in monorepo and will eventually remove it from styled-ppx. That's why Css isn't at pair with CssJs (where some bugfix got fixed only in CssJs)

PR #499, opened 2024-08-07
Make property optional on the transtion make function

zakybilfagih:
We make it optional in the make function instead adding the default value via the PPX.

Need to do this with other shorthand properties as well, so that the API is consistent.


davesnx (approved):
Much better to align APIs, good job!

PR #502, opened 2024-08-07
Cleanup grid properties

zakybilfagih:
We make the length and cascading lax here so that they can coerce to other type when interpolating.

### DIFF
--- a/packages/runtime/native/shared/Css_types.ml
+++ b/packages/runtime/native/shared/Css_types.ml
@@ -729,20 +729,6 @@ module TimingFunction = struct
     | #Cascading.t as c -> Cascading.toString c
 end
 
-module RepeatValue = struct
-  type t =
-    [ `autoFill
-    | `autoFit
-    | `num of int
-    ]
-
-  let toString x =
-    match x with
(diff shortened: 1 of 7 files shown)


davesnx (approved):
Make sense, the API is a bit weird thought, not much we can do about it tbh

packages/runtime/native/shared/Css_types.ml
```diff
@@ -4076,3 +4062,167 @@ module FontVariantEmoji = struct
...
+    | `subgrid -> {js|subgrid|js}
+    | #FixedSize.t as x -> FixedSize.toString x
+    | #Repeat.t as x -> Repeat.toString x
+    | #TrackSize.t as x -> TrackSize.toString x
+end
+
+module GridTemplateRows = struct
+  type t =
+    [ None.t
+    | `value of Track.t array
```
davesnx, line 4213: maybe there should be a function to construct those `values?

zakybilfagih: yeap, just realized that. lineNames, subgrid, etc

PR #504, opened 2024-08-11
Handle var() globally

### LINKED ISSUE #417: Support CSS variables inside calc
CSS Variables aren’t well supported on [[Parser.re](http://parser.re/)](http://Parser.re) and some properties can contain CSS variables.

It’s a similar issue as <interpolation> where it can be in a lot of places, and the CSS Spec doesn’t explain this in the BNF.

zakybilfagih:
Closes #417, we handle this the same way as cascading values.
If we detect the use of var(), it will be unsafe and the value will be rendered as string.


zakybilfagih:
I see your point, I think we can put them exactly the same place as we put <interpolation> on them Parser grammar.
I also have some thoughts on how we place <interpolation> on the grammar, will create an abt that.

PR #505, opened 2024-08-14
Update CHANGES.md

zakybilfagih:


CHANGES.md
```diff
@@ -1,5 +1,12 @@
 # Changes
 
+## 0.59.1
```
davesnx, line 3: If we have a bunch of breaking, let's do 0.60

PR #506, opened 2024-08-15
Fix calc function

zakybilfagih:


packages/runtime/native/shared/Css_types.ml
```diff
@@ -190,7 +189,7 @@ module Length = struct
     | `add of calc_value * calc_value
     | `sub of calc_value * calc_value
     | `mult of calc_value * calc_value
-    | `div of calc_value * float
+    | `div of calc_value * calc_value
```
davesnx, line 192: heads up, this doesn't seem right. You can't divide by something that isn't a number.

davesnx: Ok, it's a bit more complex than this:

```
Current implementations require that for the * and / operators, one of the operands has to be unitless. For /, the right operand must be unitless. For example font-size: calc(1.25rem / 1.25) is valid but font-size: calc(1.25rem / 125%) is invalid.
```

https://developer.mozilla.org/en-US/docs/Web/CSS/calc

zakybilfagih: Yeah the formal grammar is a bit misleading.
```
<calc-product> = <calc-value> [ [ '*' |  '/' ] <calc-value>]*
```
 maybe we can have <unitless-calc-value>

zakybilfagih: maybe later, i think float should be sufficient for now

packages/ppx/test/css-support/color-module.t/run.t
```diff
@@ -146,5 +146,5 @@ If this test fail means that the module is not in sync with the ppx
   CSS.unsafe({js|columnRuleColor|js}, {js|#000000FF|js});
   CSS.unsafe({js|columnRuleColor|js}, {js|rebeccapurple|js});
   
-  CSS.color(`rgba((0, 0, 0, `num(1.))));
+  CSS.color(`rgba((0, 0, 0, `calc(`num(1.)))));
```
davesnx, line 149: I don't think this is correct, calc shouldn't leak here

zakybilfagih: So calc(1) should evaluate to 1?

davesnx: Nevermind, I read the diff wrong. I thought calc wasn't on the value.

PR #507, opened 2024-08-20
Add `nix.flake` for easy `devshell`

frekw:


flake.nix
```diff
@@ -0,0 +1,132 @@
...
+      type = "git";
+      url = "https://github.com/ml-in-barcelona/quickjs.ml.git";
+      submodules = true;
+      flake = false;
+    };
+
+    server-reason-react = {
+      url = "github:ml-in-barcelona/server-reason-react";
+      flake = false;
+    };
```
davesnx, line 29: server-reason-react depends on quickjs, do we need to annotate both here?

davesnx:
Hey @frekw, you are probably the only one using nix with styled-ppx. If you want to push this forward, I suggest configuring some CI or steps I can run to ensure the workflow remains working, otherwise... It would never work

PR #508, opened 2024-08-20
Fix Reason compilation error

frekw:
Builds on top of #507.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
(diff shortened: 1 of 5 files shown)


frekw:
Fixed by removing [this Reason patch](https://github.com/nix-ocaml/nix-overlays/blob/master/ocaml/reason/patches/0001-rename-labels.patch) from my build via an overlay:


```
ocamlPackages = super.ocaml-ng.ocamlPackages_5_1.overrideScope' (
  oself: osuper: {
    reason = osuper.reason.overrideAttrs (o: {
      patches = [ ];
    });
}```

PR #511, opened 2024-09-13
Use buffer to build stylesheet

### LINKED ISSUE #503: Remove all usages of Printf in native runtime
Similar to what has been done in html_of_jsx or server-reason-react, we don't need to allocate the strings while concat... since it's just the output 

### References
https://github.com/ml-in-barcelona/server-reason-react/pull/156
https://github.com/davesnx/html_of_jsx/pull/18

zakybilfagih:
Resolve #503 wohoo

### DIFF
--- a/packages/runtime/native/CSS.ml
+++ b/packages/runtime/native/CSS.ml
@@ -35,20 +35,22 @@ module Array = struct
   let flatten a = Array.concat (Array.to_list a)
 end
 
-let render_declaration rule =
+let render_declaration out rule =
   match rule with
   (* https://emotion.sh/docs/labels should be ignored on the rendering *)
-  | Rule.Declaration ("label", _value) -> None
+  | Rule.Declaration ("label", _value) -> ()
   | Rule.Declaration (property, value) ->
-    Some (Printf.sprintf "%s: %s;" property value)
-  | _ -> None
+    Buffer.add_string out property;
+    Buffer.add_string out ": ";
+    Buffer.add_string out value;
+    Buffer.add_char out ';';
+    Buffer.add_char out ' '
+  | _ -> ()
 
-let render_declarations rules =
-  rules
-  |> Array.map ~f:Autoprefixer.prefix
-  |> Array.flatten
-  |> Array.filter_map ~f:render_declaration
-  |> Array.join ~sep:" "
+let render_declarations out rules =
+  let ls = rules |> Array.map ~f:Autoprefixer.prefix |> Array.flatten in
+  ls |> Array.iter ~f:(render_declaration out);
+  if Array.length ls > 0 then Buffer.truncate out (Buffer.length out - 1)
 
 let contains_at selector = String.contains selector '@'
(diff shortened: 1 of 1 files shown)


packages/runtime/native/CSS.ml
```diff
@@ -35,20 +35,22 @@ module Array = struct
...
-  rules
-  |> Array.map ~f:Autoprefixer.prefix
-  |> Array.flatten
-  |> Array.filter_map ~f:render_declaration
-  |> Array.join ~sep:" "
+    Buffer.add_string out property;
+    Buffer.add_string out ": ";
+    Buffer.add_string out value;
+    Buffer.add_char out ';';
+    Buffer.add_char out ' '
```
davesnx, line 47: Super nit, but `add_string "; "`

packages/runtime/native/CSS.ml
```diff
@@ -35,20 +35,22 @@ module Array = struct
...
+  | _ -> ()
+
+let render_declarations out rules =
+  let ls = rules |> Array.map ~f:Autoprefixer.prefix |> Array.flatten in
```
davesnx, line 51: I have seen "ls" in a lot of places and I'm rather confused, could you make a better name?

packages/runtime/native/CSS.ml
```diff
@@ -35,20 +35,22 @@ module Array = struct
...
+  ls |> Array.iter ~f:(render_declaration out);
+  if Array.length ls > 0 then Buffer.truncate out (Buffer.length out - 1)
```
davesnx, line 53: Nice trick

packages/runtime/native/CSS.ml
```diff
@@ -300,68 +302,81 @@ let resolve_selectors rules =
...
-  let push = Buffer.add_string buff in
-  let rule_to_string rule =
-    match rule with
-    | Rule.Declaration (property, value) ->
-      push (Printf.sprintf "%s:%s;" property value)
-    | Rule.Selector (selector, rules) ->
-      let rules = rules_to_string rules in
-      push (Printf.sprintf "%s{%s}" selector rules)
+let rules_to_string rules =
+  let out = Buffer.create 1024 in
```
davesnx, line 361: Just an idea, instead of passing the buffer everywhere, what do you think about passing a `~push` callback? Not suggesting to do the change, but rather to discuss if you see it better or worst.

davesnx (approved):
Cool PR @zakybilfagih

packages/runtime/native/CSS.ml
zakybilfagih: This could work, I guess the upside would be we are not dependent on Buffer.
Since we can push string and char, a possible solution would be using two callbacks `~pushc` `~pushs`, or maybe a polymorphic variant prop `` `string of string | `char of char `` ?

davesnx: meh, true. Not a big deal

davesnx (approved):

PR #513, opened 2024-09-30
Add unicodeRange to fontFace

pedrobslisboa:
## Description

It was missing the unicode-range property on fontFace. This PR adds it.
It also ran the `make format` 

The `unicode-range` works under this type:

```ocaml
  type t =
    [ `single of string
    (* E.g. `single "1234" U+1234 *)
    | `range of string * string
    (* E.g. `range ("1234", "5678") U+1234-U+5678 *)
    | `wildcard of string * string
    (* E.g. `wildcard ("12", "??") U+12?? *)
    ]
    array
```

@davesnx I'm not so happy with ``wildcard ("12", "??")`. Do you have a better idea?

### DIFF
--- a/packages/runtime/native/shared/Css_types.ml
+++ b/packages/runtime/native/shared/Css_types.ml
@@ -4066,13 +4066,17 @@ end
 
 module URange = struct
   type t =
-    [ `single of string | `range of string * string | `wildcard of string * string ] array
+    [ `single of string
+    | `range of string * string
+    | `wildcard of string * string
+    ]
+    array
 
   let toString (x : t) =
     Kloth.Array.map_and_join
       ~f:(function
         | `single x -> "U+" ^ x
-        | `range (x, y) -> "U" ^ x ^ "-" ^ y
+        | `range (x, y) -> "U+" ^ x ^ "-" ^ y
         | `wildcard (x, y) -> "U+" ^ x ^ y)
       ~sep:{js|, |js} x
 end
--- a/packages/runtime/native/shared/Css_types.ml
+++ b/packages/runtime/native/shared/Css_types.ml
@@ -4064,6 +4064,19 @@ module FontVariantEmoji = struct
     | #Cascading.t as c -> Cascading.toString c
 end
 
+module URange = struct
+  type t =
+    [ `single of string | `range of string * string | `wildcard of string * string ] array
+
+  let toString (x : t) =
+    Kloth.Array.map_and_join
+      ~f:(function
+        | `single x -> "U+" ^ x
+        | `range (x, y) -> "U" ^ x ^ "-" ^ y
+        | `wildcard (x, y) -> "U+" ^ x ^ y)
+      ~sep:{js|, |js} x
+end
+
 module InflexibleBreadth = struct
   type t =
     [ Length.t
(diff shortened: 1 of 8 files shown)


davesnx:
> @davesnx I'm not so happy with ``wildcard ("12", "??")`. Do you have a better idea?

what about ``wildcard ("12")` and render ?? directly?

packages/runtime/native/shared/Css_types.ml
```diff
@@ -4064,6 +4064,23 @@ module FontVariantEmoji = struct
     | #Cascading.t as c -> Cascading.toString c
 end
 
+module URange = struct
+  type t =
+    [ `single of string
```
davesnx, line 4069: I would turn strings into ints, right?

packages/runtime/test/test_styles.ml
```diff
@@ -1046,6 +1061,7 @@ let tests =
       avoid_hash_collision;
       keyframe;
       global;
+      fontFace;
```
davesnx, line 1064: Nice test

davesnx (approved):

packages/runtime/native/shared/Css_types.ml
pedrobslisboa: My mistake, doing it.

pedrobslisboa: Just tried it but it breaks the hex string_of_int:

E.g: 0x3F Becomes 63.

I can make it work with Printf.sprintf, but it breaks the CI on rescript:

```
  We've found a bug for you!
  packages/runtime/native/shared/Css_types.ml:4079:28-41

  The module or file Printf can't be found.
  - If it's a third-party dependency:
    - Did you list it in bsconfig.json?
    - Did you run `rescript build` instead of `rescript build -with-deps`
      (latter builds third-parties)?
  - Did you include the file's directory in bsconfig.json?
  ```
Do you have an alternative in mind?  Can It be added to dependency? Or it should be created a function to handle this string_of_hex_int?

Comment (davesnx) [reply packages/runtime/native/shared/Css_types.ml:4069]: oh, crap. Let's keep using strings for this.

**Comment by pedrobslisboa** [comment]
> > @davesnx I'm not so happy with ``wildcard ("12", "??")`. Do you have a better idea?
> 
> what about ``wildcard ("12")` and render ?? directly?

I wonder if U+4? matches: U+40, U+400, U+4000, U+4000... . If don't, adding `??` dynamically may not be possible.

**Comment by davesnx** [comment]
`U+4??` means include all characters in the range `U+400` to `U+4FF`.

https://drafts.csswg.org/css-fonts/#unicode-range-desc

Your design is probably right:

```
Wildcard ranges specified with ‘?’ that lack an initial digit (e.g. "U+???") are valid and equivalent to a wildcard range with an initial zero digit (e.g. "U+0???" = "U+0000-0FFF"). Wildcard ranges that extend beyond the range of Unicode codepoints are invalid. Because of this, the maximum number of trailing '?' wildcard characters is five, even though the [UNICODE-RANGE](https://www.w3.org/TR/CSS21/syndata.html#tokenization) token accepts six.
```

PR #517, opened 2024-10-22
Remove space between keyframe rules

pedrobslisboa:
## Problem

Consider the following OCaml code:

```ocaml
let loading = [%keyframe {|
  0% { background-position: 1rem 0 ; }
  100% { background-position: 0 0 ; }
|}]
```

In this example, the `loading` animation's name generated on the server is `animation-1fwlinh`, but on the client, it appears as `animation-1qwy1ey`, leading to a mismatch.

## Description

While working with Server-Side Rendering (SSR) and hydration, I encountered a hard-to-diagnose issue:

```
Warning: Prop `className` did not match. Server: "css-ygijhp css-1j1ftdp css-1tyndxa css-1nns36c css-1y1tzge" Client: "css-ygijhp css-1ip10c0 css-1tyndxa css-1nns36c css-1y1tzge"
```

Initially, I suspected the problem was related to a misused `switch%platform`. After extensive debugging, I discovered that the root cause was the difference in how animation names are generated between the server and client environments. The discrepancy arises from an extra space added between keyframe rules when processed natively, compared to the behavior in `emotion`.

The key issue lies in the following keyframe rule:

```ocaml
let loading = [%keyframe {|
  0% { background-position: 1rem 0 ; }
  100% { background-position: 0 0 ; }
|}]
```

This produces the following string for hashing: "0%{transform:rotate(0deg);} 100%{transform:rotate(-360deg);}"

As a result, the generated animation name on the server is `animation-1fwlinh`, but on the client, it is `animation-1qwy1ey`, causing the class name mismatch.

To resolve the issue, I found that removing the space between keyframe rules in the CSS resolves the mismatch:

```diff
- 0%{transform:rotate(0deg);} 100%{transform:rotate(-360deg);}
+ 0%{transform:rotate(0deg);}100%{transform:rotate(-360deg);}
```

This change ensures that the animation name is consistent between server and client: `animation-1qwy1ey`.

This slight difference in spacing affects the hash generated for the animation name, which explains the discrepancy between SSR and client-side rendering.

To make sure it was the behavior I checked the emotion code and there was no space for keyframes as we can see on those files: 
https://github.com/emotion-js/emotion/blob/e310c6e6a9f0b4f55030831ef65cd07c7105ae01/packages/css/src/create-instance.ts#L111-L122
https://github.com/emotion-js/emotion/blob/e310c6e6a9f0b4f55030831ef65cd07c7105ae01/packages/serialize/src/index.ts#L385-L470
On loop, like ours, there is no space added: https://github.com/emotion-js/emotion/blob/e310c6e6a9f0b4f55030831ef65cd07c7105ae01/packages/serialize/src/index.ts#L418-L431

### DIFF
--- a/packages/runtime/native/CSS.ml
+++ b/packages/runtime/native/CSS.ml
@@ -412,8 +412,7 @@ end
 
 let keyframes_to_string keyframes =
   let buffer = Buffer.create 1024 in
-  Array.iteri keyframes ~f:(fun i (percentage, rules) ->
-      if i > 0 then Buffer.add_char buffer ' ';
+  Array.iter keyframes ~f:(fun (percentage, rules) ->
       Buffer.add_string buffer (string_of_int percentage);
       Buffer.add_string buffer "%{";
       Buffer.add_string buffer (rules_to_string rules);


davesnx:
Are no tests affected by this space?

pedrobslisboa:
> Are no tests affected by this space?

Somehow, no test broke 🤔

davesnx:
Then we should add a test for future issues

pedrobslisboa:
> Then we should add a test for future issues

@davesnx I notice that we are having this issue because the test is using the hash produced by CSS.keyframes to use it on the match:

https://github.com/pedrobslisboa/styled-ppx/blob/59246af260df5417ae0c17669dd401f04405d298/packages/runtime/test/test_styles.ml#L649-L678
```ocaml
let style_tag =
  test "style_tag" @@ fun () ->
  CSS.global [| CSS.selector "html" [| CSS.lineHeight (`abs 1.15) |] |];
  let animationName =
    CSS.keyframes
      [|
        0, [| CSS.transform (`rotate (`deg 0.)) |];
        100, [| CSS.transform (`rotate (`deg (-360.))) |];
      |]
  in
  let animationName = Css_types.AnimationName.toString animationName in
  let animationNameHash =
    String.sub animationName 10 (String.length animationName - 10)
  in
  let classname = CSS.style [| CSS.display `block |] in
  let classname_hash = String.sub classname 4 (String.length classname - 4) in
  let css = CSS.style_tag () |> ReactDOM.renderToString in
  let _ = CSS.flush () in
  (* This is the hash of the global styles, since we don't capture it with global, we inline it for the test *)
  let global_hash = "18zdck7" in
  assert_string css
    (Printf.sprintf
       "<style data-emotion=\"css %s %s %s\" data-s>html{line-height:1.15;} \
        @keyframes %s { 0%% { -webkit-transform: rotate(0deg); -moz-transform: \
        rotate(0deg); -ms-transform: rotate(0deg); transform: rotate(0deg); } \
        100%% { -webkit-transform: rotate(-360deg); -moz-transform: \
        rotate(-360deg); -ms-transform: rotate(-360deg); transform: \
        rotate(-360deg); } } .%s { display: block; }</style>"
       global_hash animationNameHash classname_hash animationName classname)
```

So, the test doesn't break because it does not know the hash.
We could have it as a hardcode or a snapshot to ensure that it changes when necessary. Do you have another idea?

PR #518, opened 2024-10-30
Upgrade to OCaml 5.2

feihong:


package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "@davesnx/styled-ppx",
-  "version": "0.54.1",
+  "version": "0.56.0",
```
feihong, line 3: @davesnx This changed on its own without me actually editing it, is it OK?

davesnx: weird, but it's fine

Issue #521, opened 2024-11-11
Hash of some special char diferent on native and client


pedrobslisboa:
## Description 

When hashing some special chars, we create different values on native than client.
For know, I'm thinking the issue is with chars Alt 130 and above: https://www.special-characters.com/

For example, this string `label:styleWithContent;content:"â"` hashes to `106y846` but it is supposed to be `xwt3po`

PR #522, opened 2024-11-13
Improve murmur2 to behave as JS on special chars

pedrobslisboa:
## Description

The murmur2 code doesn't fit well on special chars (#521)

```diff
   $ ./compare.sh 'content: "é"'
+  Hashes do not match
+  native: corivl
+  JavaScript: nawgoz
+  [1]
   $ ./compare.sh 'content: "®"'
+  Hashes do not match
+  native: 3084ue
+  JavaScript: ufik94
+  [1]
   $ ./compare.sh 'content: "😀"'
+  Hashes do not match
+  native: cqlzcw
+  JavaScript: 1u8rdd
+  [1]
```

## How

There were some issues that without handling special char we didn't see before:

- JS strings are UTF-16 encoded, so for chars, the max length is 2, but ocaml strings are UTF-8 encoded, meaning they use a variable length (1–4 bytes);
  - Then I create the `get_utf16_char_codes`, and instead of handling with string on the murmur, we hold with a list of utf16 char codes.
- The emotion murmur has trick cases, where it handles Int in some instances and Int32 in others. So, I changed to primary Int and used Int32 for some specific cases.

It probably worth a blog post about Int and int32 on JS and utf-16


packages/murmur2/dune
```diff
@@ -1,3 +1,4 @@
 (library
  (name murmur2)
+ (libraries uutf)
```
jchavarri, line 3: Do you know if it's possible to not add a dependency? Related: https://discuss.ocaml.org/t/decoding-many-unicode-strings-with-uutf/8910/2 and https://github.com/ocaml/ocaml/pull/10710.

Relevant sentence:
> And once you can afford 4.14, I suggest you ditch uutf for the [Stdlib UTF decoders](https://github.com/ocaml/ocaml/pull/10710) which do not allocate at all.

packages/murmur2/test/hash.t
```diff
@@ -79,6 +79,34 @@
   $ ./compare.sh "z-index: 10"
   Hashes match: 7au0g0
 
+Special chars
```
jchavarri, line 82: I asked chatgpt for more examples of tests (I don't think the "long strings" one will compile 😓 but still useful):

Adding more targeted tests can ensure that edge cases and potential boundary issues are well-covered. Here are some specific examples of tests to add:

### 1. **Boundary Tests for UTF-16 Code Units**
   - These tests would ensure the code correctly handles characters around the boundary between single UTF-16 code units and surrogate pairs.

   ```ocaml
   $ ./compare.sh 'content: "\u{FFFF}"' (* Last code point within the BMP *)
   Hashes match: expected_hash

   $ ./compare.sh 'content: "\u{10000}"' (* First code point requiring surrogate pairs *)
   Hashes match: expected_hash
   ```

### 2. **Surrogate Pair Ranges**
   - Include specific tests for both high and low surrogates to verify correct handling of surrogate pairs, particularly ensuring they combine correctly.

   ```ocaml
   $ ./compare.sh 'content: "\u{D800}"' (* High surrogate, standalone (should not occur normally) *)
   $ ./compare.sh 'content: "\u{DC00}"' (* Low surrogate, standalone (should not occur normally) *)
   ```

### 3. **Multi-Character Strings with Mixed BMP and Non-BMP Characters**
   - Testing strings that mix BMP characters (e.g., ASCII, accented letters) and characters that require surrogate pairs helps confirm consistent encoding and hashing across different character ranges.

   ```ocaml
   $ ./compare.sh 'content: "a𐍈z"' (* ASCII, non-BMP character, ASCII *)
   Hashes match: expected_hash

   $ ./compare.sh 'content: "é😀中"' (* Accented character, emoji, CJK character *)
   Hashes match: expected_hash
   ```

### 4. **Edge Cases for Non-Printable ASCII and Control Characters**
   - Testing with control characters (e.g., newline, tab) ensures they’re handled appropriately in UTF-16 encoding.

   ```ocaml
   $ ./compare.sh 'content: "Line1\nLine2"' (* Newline in string *)
   Hashes match: expected_hash

   $ ./compare.sh 'content: "Tab\tSpace"' (* Tab character *)
   Hashes match: expected_hash
   ```

### 5. **Long Strings**
   - Test long strings to validate that your code handles lengthy input without performance issues or memory problems, especially important for a hashing algorithm.

   ```ocaml
   $ ./compare.sh 'content: "a" ^ 1000' (* Repeat a single character 1000 times *)
   Hashes match: expected_hash

   $ ./compare.sh 'content: "😀" ^ 500' (* Repeat a surrogate pair character 500 times *)
   Hashes match: expected_hash
   ```

### 6. **Combination of Diacritical Marks and Composite Characters**
   - Certain characters can be combined with diacritical marks to form composite characters. Testing these combinations ensures accurate encoding and hashing.

   ```ocaml
   $ ./compare.sh 'content: "e\u0301"' (* 'e' + acute accent (é) *)
   Hashes match: expected_hash

   $ ./compare.sh 'content: "n\u0303o\u0301"' (* 'n' + tilde, 'o' + acute accent (ñó) *)
   Hashes match: expected_hash
   ```

### 7. **Emoji Sequences and Skin Tone Modifiers**
   - Emojis with skin tone or gender modifiers form composite characters in UTF-16, which could affect hashing.

   ```ocaml
   $ ./compare.sh 'content: "👍🏽"' (* Thumbs up with medium skin tone *)
   Hashes match: expected_hash

   $ ./compare.sh 'content: "👩‍💻"' (* Woman technologist emoji, sequence with zero-width joiner *)
   Hashes match: expected_hash
   ```

### 8. **CJK and Other Multi-Byte Characters**
   - Characters from CJK languages and other multi-byte character sets are crucial to test as they are common in non-Latin scripts and can have varied byte-length representations in UTF-16.

   ```ocaml
   $ ./compare.sh 'content: "漢字"' (* Chinese characters *)
   Hashes match: expected_hash

   $ ./compare.sh 'content: "हिन्दी"' (* Hindi script *)
   Hashes match: expected_hash
   ```

These examples aim to cover a wide variety of characters, including edge cases for surrogate pairs, multi-byte characters, and sequences. They will help validate that the hashing algorithm performs consistently with JavaScript’s UTF-16 expectations across diverse inputs.

jchavarri (commented):
This is very cool! ❤️ My only concern is if `uutf` lib is needed or this can be solved with OCaml stdlib.

packages/murmur2/dune
pedrobslisboa: Just remove the `uutf` adding a few lines of code. Thank you, @jchavarri

davesnx (approved):
Woa

packages/murmur2/murmur2.ml
```diff
@@ -2,75 +2,132 @@ let char_map_36 = "0123456789abcdefghijklmnopqrstuvwxyz"
...
-  let h = ref 0l in
+  let mix_int k = Int32.to_int (Int32.mul (Int32.of_int k) (Int32.of_int m)) in
 
-  (* Mix 4 bytes at a time into the hash *)
+  (* Initialize the hash *)
+  let h = ref 0 in
   let i = ref 0 in
-  let len = ref (String.length str) in
+
+  let utf8_values = get_utf16_char_codes str in
```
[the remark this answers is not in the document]
davesnx: Maybe just try/catch it? or more important what happens when this fails in murmurjs?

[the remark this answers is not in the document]
pedrobslisboa: Hey @davesnx, thank you for this comment.

`mumurjs` does not raise but it indeed handles malformed chars, and the handle with �

So I updated the `get_utf16_char_codes` to not raise and to return the necessary amount.

I also added a test for it, I didn't like the way I made this test; too many files, but I needed ideas.

path/to/file
```diff
@@ -0,0 +1 @@
+
```
jchavarri, line 1: ✂️

packages/murmur2/test/cli_murmur2.ml
```diff
@@ -1,8 +1,24 @@
-let () =
-  if Array.length Sys.argv <> 2 then (
-    Printf.eprintf "Usage: %s <input_string>\n" Sys.argv.(0);
-    exit 1);
-
+if Array.length Sys.argv <> 2 then (
+  Printf.eprintf "Usage: %s <input_string>\n" Sys.argv.(0);
+  exit 1)
+else if Sys.argv.(1) = "--mal-formed" then (
```
jchavarri, line 4: Is this flag used?

packages/murmur2/test/murmur2.js
jchavarri, line 1: This module used to have some comments pointing to original implementations. it seems they were removed in https://github.com/davesnx/styled-ppx/commit/284759a71f276eebd753f3de5535beb106f7e939 for some reason (I found them useful for anyone contributing to the code).

path/to/file
pedrobslisboa: 🤔

packages/murmur2/test/cli_murmur2.ml
pedrobslisboa: Nop, it was a previous attempt to create the tests for malformed, forgot to delete it.

packages/murmur2/test/murmur2.js
pedrobslisboa: It's a good point. I don't know why it was removed, but I included a comment to this file:
```js
// This code is copied from: https://github.com/emotion-js/emotion/blob/f3b268f7c52103979402da919c9c0dd3f9e0e189/packages/hash/src/index.js
// to simulate the murmur2 hash function in emotion.
```

packages/murmur2/test/hash.t
```diff
@@ -79,9 +79,90 @@
...
+  Hashes match: j9pabx
+
+The following two rule sets produced the same hash (see https://github.com/davesnx/styled-ppx/pull/376)
   $ ./compare.sh "color:var(--alt-text--tertiary);:disabled{color:var(--alt-text--tertiary);}:hover{color:var(--alt-text--primary);}"
   Hashes match: 66bc4u
   $ ./compare.sh "display:flex;:before, :after{content:'';flex:0 0 16px;}"
   Hashes match: ab0yh7
+
+Malformed UTF-8
+  $ ./compare-malformed.sh
```
davesnx, line 167: Just curious, why malformed is another test and another sh?

PR #524, opened 2024-12-03
Update server-reason-react to latest

pedrobslisboa:
## Description

This PR is just to get the newest `server-reason-react` via pin. I'm going to update it and open the PR as soon we have it on opam.

### DIFF
--- a/styled-ppx.opam.template
+++ b/styled-ppx.opam.template
@@ -1,3 +1,7 @@
 depexts: [
   ["@emotion/css"] {npm-version = ">=11.0.0"}
 ]
+
+pin-depends: [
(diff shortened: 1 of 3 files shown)


styled-ppx.opam.template
```diff
@@ -1,3 +1,7 @@
 depexts: [
   ["@emotion/css"] {npm-version = ">=11.0.0"}
 ]
+
+pin-depends: [
+  ["server-reason-react.0.3.1" "git+https://github.com/ml-in-barcelona/server-reason-react#ea16c107cd41a2e9d01dff358239eec37046951f"]
```
davesnx, line 6: You might need to add it on the "pin" command in the makefile for the CI to work

PR #530, opened 2024-12-14
Parser render module

zakybilfagih:


packages/css-property-parser/lib/Standard.re
```diff
@@ -253,16 +257,39 @@ let interpolation = {
...
+    fun
+    | IDENT(value) => {
+        switch (value) {
+        | "only"
+        | "not"
+        | "and"
+        | "or"
+        | "layer" =>
+          Error([
+            "'only', 'not', 'and', 'or', and 'layer' are invalid media types",
```
davesnx, line 271: yo, give me a proper error message. 

can you do a `as used` and Printf.sprintf "%s is not a valid media type" used?

packages/css-property-parser/lib/Standard.re
```diff
@@ -253,16 +257,39 @@ let interpolation = {
...
+let container_name = {
+  open Rule.Let;
+  let.bind_match name = custom_ident;
+  let value = {
+    switch (name) {
+    | "none"
+    | "and"
+    | "not"
+    | "or" =>
+      Error(["'none', 'and', 'not', and 'or' are invalid container names"])
```
davesnx, line 288: How's possible that we don't catch those at parse time?

packages/parser/lib/Lexer.re
```diff
@@ -542,22 +532,28 @@ let handle_tokenizer_error = lexbuf => {
 };
 
 let skip_whitespace = ref(false);
+let prev_skip_whitespace = ref(false);
```
davesnx, line 535: whoa, I would need an explanation on why previous is needed, when we have skip_ws

davesnx: If both booleans are needed, I would go with a single ref and a few variants, wdyt?

Inside_media | Inside_selector | Outside_x | etc..

packages/parser/test/Render_test.t
```diff
@@ -0,0 +1,4 @@
+  $ alias run='./Render_test.exe'
+
+  $ cat ./test.css | run
```
davesnx, line 3: why the alias?

packages/parser/test/Render_test.t
```diff
@@ -0,0 +1,4 @@
...
+  .some-grid{grid:[line1] minmax(20em, max-content) / auto-flow dense 40%;}.a,href > h1,h3[data-foo*=bar]{&::after,href > h1,&:is(h1),&:has(+ :is(h3)){padding:20px 30px calc(20px+30px*(20 * 400/2));}background-image:radial-gradient(circle, #0000 45%, #000f 48%), radial-gradient(ellipse farthest-corner, #fc1c14 20%, #cf15cf 80%);grid-template-columns:[line-name1] 100px [line-name2] repeat(auto-fit, [line-name3 line-name4] 300px) ;}abbr{color:chocolate;}@media (hover: hover) {abbr:hover{color:limegreen;transition-duration:1s;}}@media not all and (hover: hover) {abbr::after{content:" (" attr(title) ")";}}
```
davesnx, line 4: Can we have those tests as "idempotent" and parse-print-parse and compare?

I would prefer to keep separate "features" as separate tests, but a big one is also cool

davesnx: maybe the "big" one can be a small tailwind generated file?

packages/ppx/src/Css_to_runtime.re
```diff
@@ -110,23 +110,29 @@ let rec render_at_rule = (~loc, at_rule: at_rule) => {
...
-      /* TODO: String.trim is a hack around, but a media query should be all clean. */
-      Ok(source_code_of_loc(ident_loc) |> String.trim)
-    /* In any other case, we believe on the source_code and transform it to string. This is unsafe */
-    | _whatever => Ok(source_code_of_loc(at_rule_loc) |> String.trim)
-    };
+    let prelude = source_code_of_loc(at_rule_prelude_loc) |> String.trim;
+    Css_property_parser.Parser.parse(
+      Css_property_parser.Parser.media_query_list,
+      prelude,
+    )
```
davesnx, line 119: Oh, I like this parsing

packages/ppx/src/Css_to_runtime.re
```diff
@@ -110,23 +110,29 @@ let rec render_at_rule = (~loc, at_rule: at_rule) => {
...
 
   switch (parse_condition) {
-  | Error(error_expr) => error_expr
+  | Error(error_msg) =>
+    Error.expr(
+      ~loc=
+        Styled_ppx_css_parser.Parser_location.update_pos_lnum(
+          at_rule_prelude_loc,
+          loc,
+        ),
```
davesnx, line 133: When have you seen this to be needed? Can add a test for it? ^^

packages/ppx/test/native/At_rule_test.re
```diff
@@ -83,25 +83,20 @@ let media_query_tests = [
...
-    "(screen, print)",
-    [%expr [%cx "@media (screen, print) {}"]],
-    [%expr CSS.style([|CSS.media({js|(screen, print)|js}, [||])|])],
-  ),
-  (
-    "(screen and (color))",
-    [%expr [%cx "@media (screen and (color)) {}"]],
-    [%expr CSS.style([|CSS.media({js|(screen and (color))|js}, [||])|])],
+    "((screen) and (color))",
+    [%expr [%cx "@media ((screen) and (color)) {}"]],
```
davesnx, line 87: This looks incorrect, the right way would be 

```
@media screen and (color) {
  color: blue;  
}
```

Parsing is: `@media [media-type] and ([media-feature]) { }`

packages/ppx/src/Css_to_runtime.re
zakybilfagih: yeah super slick

packages/css-property-parser/lib/Standard.re
zakybilfagih: hahaha, will fix this

packages/css-property-parser/lib/Standard.re
zakybilfagih: media_type has the value <ident> but the spec specifically asked to exclude those values

packages/parser/lib/Lexer.re
zakybilfagih: oof forgot to refactor this specific hack hehe.
when we are inside a brace, the skip_whitespace is set to true and the whitespace inside values (<property>: <values>) is skipped. i want to enable whitespace when we are seeing `:` and bring skip_whitespace to its previous value when we are seeing `;`

i like the variant idea though, will try to implement it!

packages/parser/test/Render_test.t
zakybilfagih: haha i copied the test from andrey's repo and forgot to change hehe

packages/parser/test/Render_test.t
zakybilfagih: yup, will be adding the parse-print-parse test asap

zakybilfagih: > maybe the "big" one can be a small tailwind generated file?

yup great idea

packages/ppx/src/Css_to_runtime.re
zakybilfagih: this keeps the cnum of the first location but offset the lnum with the second loc. i've tried using intersection but this seems to work well to highlight the correct error location.

packages/ppx/test/native/At_rule_test.re
zakybilfagih: i think we tested that already, this should resolve to this https://csstree.github.io/docs/syntax/#Type:media-query-list
wat

packages/parser/lib/Resolve.re
```diff
@@ -0,0 +1,518 @@
+open Ast;
+
+let rec contains_ampersand = (selector: selector) => {
+  switch (selector) {
+  | SimpleSelector(Ampersand) => true
+  | ComplexSelector(Selector(selector)) => contains_ampersand(selector)
+  | ComplexSelector(Combinator({left: selector_left, right})) =>
+    contains_ampersand(selector_left)
```
davesnx, line 8: did we migrate tests for all of this?

packages/parser/test/Lexer_test.re
```diff
@@ -47,17 +47,17 @@ let success_tests =
...
-    ({|and |}, [MEDIA_QUERY_OPERATOR("and ")]),
-    ({|or |}, [MEDIA_QUERY_OPERATOR("or ")]),
-    ({|all|}, [ALL_MEDIA_TYPE("all")]),
-    ({|screen|}, [SCREEN_MEDIA_TYPE("screen")]),
-    ({|print|}, [PRINT_MEDIA_TYPE("print")]),
+    ({|and |}, [IDENT("and"), WS]),
+    ({|or |}, [IDENT("or"), WS]),
+    ({|all|}, [IDENT("all")]),
+    ({|screen|}, [IDENT("screen")]),
+    ({|print|}, [IDENT("print")]),
```
davesnx, line 60: absolute beauty

packages/parser/test/Parser_test.re
```diff
@@ -34,28 +34,29 @@ let error_tests_data =
...
-    (
-      /* whitespace must follow `and` */
-      "@media (not (color)) and(hover) {}",
-      "Parse error while reading token 'and(' on line 1 at position 20",
-    ),
-    (
-      /* space between < and = is invalid */
-      "@media (width < = 33px) {}",
-      "Parse error while reading token '=' on line 1 at position 15",
+      "Parse error while reading token '' on line 1 at position 7",
```
davesnx, line 37: weird error message; it should at least render the `$`.

packages/parser/test/Parser_test.re
```diff
@@ -34,28 +34,29 @@ let error_tests_data =
...
     ),
+    //(
```
davesnx, line 39: what's the reason for those comments?

packages/parser/lib/Render.re
```diff
@@ -0,0 +1,191 @@
...
+    Printf.sprintf("%s(%s)", fst(name), body);
+  | Hash(string) => "#" ++ string
+  | Number(n) => n
+  | Unicode_range(string) => string
+  | Float_dimension((a, b)) => Printf.sprintf("%s%s", a, b)
+  | Dimension((a, b)) => Printf.sprintf("%s%s", a, b)
+  | Variable(variable) => render_variable(variable)
+  | Selector(v) => render_selector_list(v)
+  };
+};
```
davesnx, line 191: Can you remove the prefix on all the functions `render_*`? When calling it should be as simpler as Render.component_value.

Also, can you add a Render.rei where it exports only the same methods as the parser expose?

%start <stylesheet> stylesheet
%start <rule_list> declaration_list
%start <declaration> declaration
%start <rule_list> keyframes

packages/parser/test/Render_test.t
```diff
@@ -0,0 +1,245 @@
+  $ alias run='./Render_test.exe'
```
davesnx, line 1: can you run the thing directly instead of this alias?

packages/parser/test/dune
```diff
@@ -13,3 +14,11 @@
  (deps Runner.exe)
  (action
   (run %{deps})))
+
+(executable
+ (name Render_test)
```
davesnx, line 19: can you add a cram test that ensures the render is idempotent?

`Parse -> Print -> Parse and check equality`

This way we ensure no data is lost

packages/ppx/src/Css_to_runtime.re
```diff
@@ -110,23 +110,29 @@ let rec render_at_rule = (~loc, at_rule: at_rule) => {
...
-      Ok(source_code_of_loc(ident_loc) |> String.trim)
-    /* In any other case, we believe on the source_code and transform it to string. This is unsafe */
-    | _whatever => Ok(source_code_of_loc(at_rule_loc) |> String.trim)
-    };
+    let prelude = source_code_of_loc(at_rule_prelude_loc) |> String.trim;
+    Css_property_parser.Parser.parse(
+      Css_property_parser.Parser.media_query_list,
+      prelude,
+    )
+    |> Result.map(_ => prelude);
```
davesnx, line 120: can we do something better with this error here?

packages/parser/test/Parser_test.re
zakybilfagih: since we handle the prelude parsing with the property parser, all of this should pass the first parse

packages/parser/lib/Lexer.re
```diff
@@ -912,6 +894,8 @@ let consume = lexbuf => {
     };
   | ":" => Ok(COLON)
   | ";" => Ok(SEMI_COLON)
+  | "<=" => Ok(IDENT("<="))
+  | ">=" => Ok(IDENT(">="))
```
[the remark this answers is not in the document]
zakybilfagih: Since we filter out the whitespace, there is no way of telling if there is a whitespace between gt and lt with eq. So I changed the property spec from this
```
<mf-lt> = '<' '='?
<mf-gt> = '>' '='?
```
to this
```
and mf_gt = [%value.rec "'>=' | '>'"]
and mf_lt = [%value.rec "'<=' | '<'"]
```
So need to handle >= and <= on the lexer

packages/parser/test/Parser_test.re
zakybilfagih: huh yeah, dunno why this is the case

packages/parser/test/dune
zakybilfagih: implemented this on render_test.re!

packages/parser/lib/Lexer.re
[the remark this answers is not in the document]
zakybilfagih: huh i see what u mean, we can create another token and change the keyword functon on Standard.re

PR #532, opened 2024-12-24
Ampersand not resolved correctly on global

### LINKED ISSUE #531: Selector containing multiple ampersand with one in the beginning not resolved correctly on global
```
[%styled.global
  {|
  .foo {
    color: blue;
    & .bar & .baz { color: red; }
  }
|}
];
```
got
```
html{line-height:1.15;}a{}a:hover{padding:0;}.foo{}.foo .bar & .baz{padding:0;}
```

zakybilfagih:
Fixes #531

### DIFF
--- a/packages/runtime/test/test_styles.ml
+++ b/packages/runtime/test/test_styles.ml
@@ -1045,11 +1045,38 @@ let global_with_selector =
   [%global
     {| html { line-height: 1.15; }
     a { :hover { padding: 0; } }
-    .foo { & .bar & .baz { padding: 0; } }
    |}];
   let css = get_string_style_rules () in
   assert_string css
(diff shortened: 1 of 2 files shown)


davesnx (approved):
That's correct, good catch and bless you from understand the resolve_selectors function

packages/runtime/test/test_styles.ml
```diff
@@ -1045,10 +1045,11 @@ let global_with_selector =
   [%global
     {| html { line-height: 1.15; }
     a { :hover { padding: 0; } }
+    .foo { & .bar & .baz { padding: 0; } }
```
davesnx, line 1048: Can you add this as a separate test case? ^^

PR #534, opened 2024-12-25
Rule.Selector now accepts string array instead of string

### LINKED ISSUE #533: Split multiple selectors fails on functional pseudo classes
```
.foo:not(strong, b.important) {
  color: red;
}
```
outputs
```
.foo:not(strong{color:#FF0000;}b.important){color:#FF0000;}
```

We might need to change the `Rule.Selector` definition from `Selector of string * rule array` to `Selector of string array * rule array` since `String.split_on_char(',')` will not cut it

zakybilfagih:
Fixes #533

**Breaking change** to the API, now selector accepts string array instead of string

### DIFF
--- a/packages/ppx/src/Css_to_runtime.re
+++ b/packages/ppx/src/Css_to_runtime.re
@@ -166,12 +166,15 @@ and render_container_query = (~loc, at_rule: at_rule) => {
   | Error(error_expr) => error_expr
   | Ok(conditions) =>
     let query =
-      String_interpolation.transform(
-        ~attrs,
-        ~delimiter,
-        ~loc=at_rule.loc,
-        "@container " ++ conditions,
-      );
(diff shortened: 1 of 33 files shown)


davesnx (approved):
Love this, to avoid the breaking change can we make CSS.selector include the array and CSS.selectorMany accept the array as argument?

The ppx should always generate the selectorMany, wdyt?

zakybilfagih:
> Love this, to avoid the breaking change can we make CSS.selector include the array and CSS.selector* accept the array as argument?
> 
> The ppx should always generate the selector*, wdyt?
> 
> I'm unsure what's a decent name for `selector*` (maybe selectorMany? dunno, totally up to you)

yup agreed 100%, i think selectorMany should be good enuf

PR #536, opened 2025-01-14
Fix media query on nested selector

### LINKED ISSUE #535: Media query not resolved correctly on deeply nested selector
Caught this bug while writing the test for the renderer
```
let mobile = "(max-width: 767px)" in
let classname =
    [%cx
      {|
      .foo {
        .bar {
          @media $(mobile) {
            position: relative;
          }
        }
      }
    |}]
```
Received: 
```
@media (max-width: 767px) { .css-1xnecka { position: relative; } }
```

zakybilfagih:
Fixes #535

### DIFF
--- a/packages/runtime/test/test_styles.ml
+++ b/packages/runtime/test/test_styles.ml
@@ -331,17 +331,22 @@ let selector_nested_with_mq_and_declarations =
   let classname =
     [%cx
       {|
-      li {
-        list-style-type: none;
-
-        ::before {
-          position: absolute;
-          left: -20px;
-          content: "✓";
-        }
-
+      .a {
+        color: red;
         @media $(mobile) {
-          position: relative;
+          height: 1px;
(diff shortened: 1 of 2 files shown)


packages/runtime/test/test_styles.ml
```diff
@@ -331,28 +331,34 @@ let selector_nested_with_mq_and_declarations =
...
+            .c {
+              color: yellow;
+            }
+          }
         }
       }
     |}]
   in
   let css = get_string_style_rules () in
   assert_string css
```
davesnx, line 355: Can you keep the test and add another case instead?

davesnx (approved):
LFG

packages/runtime/test/test_styles.ml
zakybilfagih: Sure!

PR #539, opened 2025-01-24
Scrollbar Module 1 and Partial Overflow Module 3 support

### LINKED ISSUE #538: Support scrollbar-gutter

zakybilfagih:
- [x] scrollbar-color
- [x] scrollbar-width
- [x] overflow-clip-margin
- [x] scrollbar-gutter

Resolve #538

### DIFF
--- a/packages/ppx/src/Property_to_runtime.re
+++ b/packages/ppx/src/Property_to_runtime.re
@@ -4496,13 +4496,12 @@ let scrollbar_color =
     (~loc, value: Types.property_scrollbar_color) =>
       switch (value) {
       | `Auto => [%expr `auto]
-      | `Color([thumbColor, trackColor]) =>
+      | `Static(thumbColor, trackColor) =>
         [%expr
          `thumbTrackColor((
            [%e render_color(~loc, thumbColor)],
            [%e render_color(~loc, trackColor)],
          ))]
-      | `Color(_) => raise(Unsupported_feature)
       },
   );
 
--- a/packages/ppx/src/Property_to_runtime.re
+++ b/packages/ppx/src/Property_to_runtime.re
@@ -2194,18 +2194,19 @@ let overflow =
 let overflow_clip_margin =
   polymorphic(
     Property_parser.property_overflow_clip_margin,
-    (~loc, (clipEdgeOrigin, margin)) =>
-    [
-      [%expr
-        CSS.overflowClipMargin2(
-          ~clipEdgeOrigin=?[%e
-            render_option(~loc, variant_to_expression, clipEdgeOrigin)
-          ],
-          ~margin=?[%e render_option(~loc, render_extended_length, margin)],
-          (),
-        )
-      ],
-    ]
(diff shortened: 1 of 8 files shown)


packages/ppx/src/Property_to_runtime.re
```diff
@@ -4462,7 +4489,21 @@ let scrollbar_base_color =
...
+    (~loc, value: Types.property_scrollbar_color) =>
+      switch (value) {
+      | `Auto => [%expr `auto]
+      | `Color([thumbColor, trackColor]) =>
+        [%expr
+         `thumbTrackColor((
+           [%e render_color(~loc, thumbColor)],
+           [%e render_color(~loc, trackColor)],
+         ))]
+      | `Color(_) => raise(Unsupported_feature)
```
davesnx, line 4504: one color looks like isn't supported (can we change the spec definition to be <color> <color> instead?

packages/runtime/native/shared/Css_types.ml
```diff
@@ -4244,3 +4244,107 @@ module GridTemplateRows = struct
...
+    type t =
+      [ Length.t
+      | Var.t
+      ]
+
+    let toString x =
+      match x with
+      | #Length.t as l -> Length.toString l
+      | #Var.t as va -> Var.toString va
+  end
```
davesnx, line 4337: Woa, didn't know!

packages/runtime/native/shared/Declarations.ml
```diff
@@ -446,6 +446,16 @@ let overflows x =
 let overflowX x = Rule.declaration ({js|overflowX|js}, Overflow.toString x)
 let overflowY x = Rule.declaration ({js|overflowY|js}, Overflow.toString x)
 
+let overflowClipMargin x =
+  Rule.declaration ({js|overflowClipMargin|js}, OverflowClipMargin.toString x)
+
+let overflowClipMargin2 ?(clipEdgeOrigin = `paddingBox) ?(margin = `px 0) () =
```
davesnx, line 452: can we keep margin not being a labelled argument? (just for cuteness)

davesnx (approved):
Good job man

packages/ppx/src/Property_to_runtime.re
zakybilfagih: this path will not be reached i believe, since we are using the curly syntax ``<color>{2}`` the type will always be a list

packages/runtime/native/shared/Declarations.ml
zakybilfagih: yeah better remove the dangling unit

packages/runtime/native/shared/Css_types.ml
zakybilfagih: 🤓

packages/ppx/src/Property_to_runtime.re
davesnx: Can you add a comment (it's easy to discover in the future but faster with a small //)

PR #544, opened 2025-03-06
Refactor `CSS.FontFamilyName.t`: Add Union Types for Unquoted Generic Families

mikaeljan:
This PR updates the type definition for `CSS.Types.FontFamilyName.t` to allow using unquoted generic font family names directly in the code. Currently, only a string is accepted, forcing generic names like "monospace" to be quoted, which isn’t ideal for specifying static defaults.

### DIFF
--- a/packages/ppx/src/Property_to_runtime.re
+++ b/packages/ppx/src/Property_to_runtime.re
@@ -2442,19 +2442,27 @@ let hanging_punctuation =
 
 let render_generic_family = (~loc) =>
   fun
-  | `Cursive => [%expr "cursive"]
-  | `Fantasy => [%expr "fantasy"]
-  | `Monospace => [%expr "monospace"]
-  | `Sans_serif => [%expr "sans-serif"]
-  | `Serif => [%expr "serif"]
-  | `_apple_system => [%expr "-apple-system"];
+  | `Cursive => [%expr `cursive]
+  | `Fantasy => [%expr `fantasy]
+  | `Monospace => [%expr `monospace]
+  | `Sans_serif => [%expr `sans_serif]
+  | `Serif => [%expr `serif]
+  | `System_ui => [%expr `system_ui]
+  | `Ui_serif => [%expr `ui_serif]
+  | `Ui_sans_serif => [%expr `ui_sans_serif]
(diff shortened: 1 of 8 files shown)


packages/ppx/src/Property_to_runtime.re
```diff
@@ -2442,19 +2442,27 @@ let hanging_punctuation =
...
+  | `Serif => [%expr `serif]
+  | `System_ui => [%expr `system_ui]
+  | `Ui_serif => [%expr `ui_serif]
+  | `Ui_sans_serif => [%expr `ui_sans_serif]
+  | `Ui_monospace => [%expr `ui_monospace]
+  | `Ui_rounded => [%expr `ui_rounded]
+  | `Emoji => [%expr `emoji]
+  | `Math => [%expr `math]
+  | `Fangsong => [%expr `fangsong]
+  | `_apple_system => [%expr `custom("-apple-system")];
```
davesnx, line 2458: apple-system isn't custom isn't it?

I believe is the old `system-ui`, but it shouldn't have the quotes

packages/runtime/native/shared/Css_types.ml
```diff
@@ -3941,13 +3941,43 @@ module FontFace = struct
...
+    | `fantasy
+    | `system_ui
+    | `ui_serif
+    | `ui_sans_serif
+    | `ui_monospace
+    | `ui_rounded
+    | `emoji
+    | `math
+    | `fangsong
+    | `custom of string
```
davesnx, line 3958: Did you thought about another name rather than "custom"? 
`quoted` or `string` comes to mind and feels cleaner. This is rather a nitpick

davesnx (approved):
LGTM

Thanks for doing this

packages/runtime/native/shared/Css_types.ml
mikaeljan: `named`  or `specific` came to mind but will stick to `quoted`

packages/ppx/src/Property_to_runtime.re
mikaeljan: you are right, I thought it was a font-name

PR #551, opened 2025-03-13
Support interpolation on url

zakybilfagih:


packages/css-property-parser/lib/Parser.re
```diff
@@ -141,7 +141,7 @@ and calc_value = [%value.rec
 and cf_final_image = [%value.rec "<image> | <color>"]
 and cf_mixing_image = [%value.rec "[ <extended-percentage> ]? && <image>"]
 and class_selector = [%value.rec "'.' <ident-token>"]
-and clip_source = [%value.rec "<url>"]
+and clip_source = [%value.rec "<url-with-interp>"]
```
davesnx, line 144: i usually name things with interp "extended", would you prefer to keep -with-interp or -extended? 

both looks good to me, but want to align here

packages/css-property-parser/lib/Parser.re
```diff
@@ -1974,6 +1974,7 @@ and symbol = [%value.rec "<string> | <image> | <custom-ident>"]
 and target = [%value.rec
   "<target-counter()> | <target-counters()> | <target-text()>"
 ]
+and url_with_interp = [%value.rec "<url> | url( <interpolation> )"]
```
davesnx, line 1977: and from the other point of view, when url isn't going to be interpolatable? maybe do a url with interpolation and call it url? wdyt?

davesnx (approved):
Aside from names inside Parser.re, LGTM

packages/css-property-parser/lib/Parser.re
zakybilfagih: i think we can use the -no-interp suffix naming? like we did for transition and animation

Issue #427, opened 2024-03-07
Umbrella issue for the CSS Support matrix


davesnx:
This is the overview of styled-ppx CSS support (as a detailed view of https://styled-ppx.vercel.app/css-support). Currently CSS 3 is almost there, but CSS lately has been adding a bunch of features that we don't support yet, but are simple to add.

## Missing properties or values under [Parser.re](http://parser.re/)

Our CSS parser fails at parsing this:

- [ ] https://github.com/davesnx/styled-ppx/issues/413
- [x] https://github.com/davesnx/styled-ppx/issues/417
- [ ] https://github.com/davesnx/styled-ppx/issues/418
- [ ] *CSS Color Module Level 5*
- [ ] *`width: attr(data-px)`*
- [ ] *Environment Variables Level 1*
- [x] *CSS Grid Layout Module Level 3*
- [ ] *`clip-path: circle()`*
- [x] *`mask-position: bottom 10px right`*
- [ ] *`mask-border: url(image.png) 10px`*
- [x] `@container` https://github.com/davesnx/styled-ppx/issues/422
- [ ] `@layer`

## Unsupported properties (code-gen)

Parsing works fine, but code-gen for those properties always generates `Css.unsafe("property", "value")`. So, interpolation doesn't work in those cases and there's risks of introducing errors at runtime while using those properties.

- [ ] https://github.com/davesnx/styled-ppx/issues/443

This task (https://github.com/davesnx/styled-ppx/issues/426) would automatically remove all the effort of going one by one manually 

- [ ] Parser.property_image_resolution
- [x] Parser.property_image_orientation
- [x] Parser.property_image_rendering
- [x] Parser.property_border_image_source
- [x] Parser.property_border_image_slice
- [x] Parser.property_border_image_width
- [x] Parser.property_border_image_outset
- [x] Parser.property_border_image_repeat
- [x] Parser.property_border_image
- [x] Parser.property_overflow_clip_margin
- [x] Parser.property_overflow_inline
- [x] Parser.property_block_ellipsis
- [x] Parser.property_max_lines
- [x] Parser.property_continue
- [x] Parser.property_tab_size
- [x] Parser.property_line_break
- [x] Parser.property_hyphens
- [x] Parser.property_text_align_all
- [x] Parser.property_text_align_last
- [x] Parser.property_text_justify
- [ ] Parser.property_hanging_punctuation
- [x] Parser.property_font_stretch
- [x] Parser.property_font_size_adjust
[...]

Issue #413, opened 2024-01-30
Support for all CSS functions


davesnx:
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Functions

PR #579, opened 2026-07-19
Accept `clamp()` in length properties

Atlas07:
## Accept `clamp()` in length properties

`clamp()` was registered as a CSS function but never listed in the `extended-length` value rule — so `width: clamp(1rem, 2vw, 3rem)` failed to validate, even though `min()` / `max()` / `calc()` worked.

### Fix
Added `<clamp()>` to `extended-length` and threaded the `Function_clamp` variant through both renderers and the runtime type, mirroring `min()`/`max()`. 
One shared rule → every length property gets it (`width`, `height`, `padding`,`margin`, `font-size`, `gap`, `inset`, …).

Scoped to `extended-length`, which covers essentially all real-world usage.
Percentages/angles/times can follow up. Relates to #413 / #427.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## 0.62.0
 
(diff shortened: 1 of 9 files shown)


davesnx (approved):
Perfect

PR #597, opened 2026-09-18
Record the dune library name in the css.config wire attribute

davesnx:
The PPX now reads the `library-name` cookie that dune passes to every ppx run inside a `(library ...)` stanza and records it in the `[@@@css.config]` wire attribute as `("library", name)`, next to the existing `env` key. The attribute is emitted whenever a file has rules or bindings and at least one key applies; a development build without the cookie still emits nothing, so existing outputs are unchanged. The aggregator ignores the new key until the follow-up library-order PR reads it to group and order rules by owning library.

```ocaml
[@@@css.config [("env", "production"); ("library", "foo")]]
```

Evidence at f02f05a5:

- `packages/ppx/test/css-support/library-cookie.t`: with `-cookie 'library-name="foo"'` a dev build emits `[@@@css.config [("library", "foo")]]`; with `--minify` it emits `[("env", "production"); ("library", "foo")]`; without the cookie no config attribute is emitted.
- `packages/generate/test/css-config.t`: a config carrying only `library` keeps development output, so the key cannot flip production mode.

Stack: first of four PRs (dependency order in `generate`, the winner-diff script, then library grouping that consumes this key).

Risk: two-way door. A revert restores the production-only emission; until then development builds inside library stanzas carry one extra config entry that no consumer reads.


davesnx:

PR #599, opened 2026-09-18
Add css-winner-diff script to compare cascade winners between stylesheets

davesnx:
`scripts/css-winner-diff.ts OLD.css NEW.css [--summary]` compares two stylesheets emitted by `styled-ppx.generate` and reports every pair of rules with the same property, specificity and at-rule wrapper but different selectors whose relative order changed, which is exactly the set of cascade ties whose winner flips. It exists to review ordering changes in the aggregator; the exit code is always 0.

Evidence at 79a3d702:

- `node scripts/css-winner-diff.ts --self-test`: all assertions pass; a 25,000-rule synthetic input completes in about 1.3 s.
- A real 8,648-rule stylesheet compared with itself: 356,464 tied pairs compared, 0 flipped.
- The monorepo's aggregated output in source order against dependency order (23,273 rules each): 2,410,908 tied pairs compared, 736,147 flipped, 0 unparsed lines.

Node 22 runs the file directly with no dependencies. Not part of the opam package.

Risk: two-way door. Tooling only; a revert removes the script and nothing else depends on it.


davesnx:
Dropped: the winner-diff script was a one-off evaluation tool with no callers in the repo. Its results stay recorded in the generate-dependency-order plan.

PR #600, opened 2026-09-18
Order aggregated CSS by library, then module dependency

davesnx:
`styled-ppx.generate` now orders rules at two levels: libraries first, then modules inside each library. A library's rules come after the rules of every library it depends on, so a consumer library's override of a dependency's atom wins by construction instead of by directory name. Inside a library the module dependency order from the previous PR applies unchanged, now scoped to that library's files.

A file's library is the `library` key the PPX records in `[@@@css.config]` from dune's `library-name` cookie; a file without the key groups with its directory. Library edges are collapsed from module references: a reference `X` that the referrer's own library defines is a module edge; otherwise it points at the one other library defining `X`, or, for wrapped libraries referenced through their alias module, at the one library whose capitalized name is `X`. Both levels reuse the same Kahn sort, so the alphabetical tiebreak and the warn-and-continue cycle policy apply at both. `--order source` still bypasses everything.

```
$ styled-ppx.generate consumer/main.ml zlib/inner.ml   # main references Zlib.Inner.x
/* This file is generated by styled-ppx, do not edit manually */
.zlib-inner{color:green;}
.consumer-main{color:blue;}
```

Evidence at ada6aede:

- New cram tests `library-order`, `library-order-wrapped`, `library-order-fallback`, `library-order-cycle`, `library-order-source-flag`; `log-flags.t` and `order-dependency.t` gain the per-library info lines; `order-cycle.t` changes because its two directories now form two groups, which is the resolution the test's own comment anticipated.
- On the Ahrefs monorepo's 5,824 post-PPX files (pinned PPX, so directory fallback, 246 groups): identical 23,274-rule set, 4.4 s against 3.7 s for the flat order; `Stack` rules still precede `Textarea`.
- `demo/melange/public/styles.css` regenerates with three rules moved and no property or value changed.

Review from `library_edge_target` and `order_by_dependency` in `packages/generate/generate.ml`. Not included: `@layer` emission, which follows as its own opt-in PR.

Risk: two-way door. A revert restores the flat module order. Until then, aggregations that mix several libraries see rules regroup by library; the winner-diff script from the previous PR lists which cascade ties flip.


davesnx:

PR #601, opened 2026-09-18
Add opt-in --layers cascade-layer emission to styled-ppx.generate

davesnx:
`styled-ppx.generate --layers` wraps each library's rules in a named cascade layer. The output starts with every `@property` and `@keyframes` rule, since those are global registrations, then one `@layer a, b, c;` statement in the emitted library order, then one `@layer <name> { ... }` block per library in that order. Layer names are the library key with characters outside `[A-Za-z0-9_-]` replaced by `_`; two keys that sanitize to the same name merge into one layer with a one-time warning. The flag is off by default and requires the dependency order; with `--order source` the generator exits 2.

```css
@property --z-base-color { syntax: '<color>'; inherits: false; initial-value: red; }
@layer z_base, a_app;
@layer z_base {
.base-widget{color:red;}
}
@layer a_app {
.app-main{color:blue;}
}
```

Why off by default: unlayered author CSS beats every layered declaration, and `!important` inverts that relation, so a consumer must layer its own resets, palettes, fonts and inline global styles first. The docs state this next to the flag.

Evidence at c7ae53c8:

- New cram tests `layers`, `layers-sanitize`, `layers-source-flag`, `layers-minified`, `layers-off`; `log-flags.t` gains one `--layers --log info` invocation.
- `layers-off.t` pins that output without the flag is byte-identical to before; `demo/melange/public/styles.css` did not change under `make build`.

Last PR of the stack; it can be dropped without affecting the ordering PRs below it.

Risk: two-way door. Opt-in flag; a revert removes it and nothing else changes.


davesnx:

PR #602, opened 2026-09-19
Mint an identity class per binding and resolve $(binding) references to it

davesnx:
Every named `[%css]` binding now mints one identity class, `cid-<hash>`, emitted ahead of its atoms, and `$(binding)` / `&.$(binding)` in selectors resolve to that one class instead of the binding's atom chain. The identity hashes the module name, submodule scope, binding name, an occurrence index when the same name repeats in one compilation unit, and the new `--namespace` flag; never a path or dune library name, so the native and melange builds of one module agree.

What changes for users:

- A reference means "carries that binding". Before, `.$(x)` expanded to `.cssA.cssB…`, which matched any element with the same declarations and had specificity (0,N,0); now it is `.cid-…` at (0,1,0).
- Empty markers `[%css {||}]` are identity-only: no rule, no `css-0-…` class. This also fixes `--minify`, where a named empty marker produced no class at all and `&.$(marker)` collapsed to `&`.
- The `[@@@css.bindings]` wire payload becomes `(longident, identity, class_string)`. The aggregator resolves sentinels to the identity and fails when two different bindings share an identity but carry different CSS, naming both input files, both longidents and the remedy: `--namespace <name>` on one library's `(pps styled-ppx …)` stanza, identically on its native and melange stanzas. Same identity with the same atoms is the ordinary "module compiled twice" case and is accepted.

```
.css-17s01vz-container.cid-12qpfnx{padding:1rem;}   /* was .css-17s01vz-container.css-1443u2l-marker.css-kbn7if-marker */
```

Labels stay in this PR; the follow-up PR removes them, which is where the dedup payoff lands (measured on the Ahrefs monorepo: 37,312 rules to 19,158, gzip 426 KB to 215 KB).

Evidence at 88e181db:

- New cram tests: `identity-class-basic`, `identity-empty-marker` (including `--minify`), `identity-local-rebinding`, `identity-namespace`, `identity-cross-path` (same module from two directories gives one identity) under `packages/ppx/test/css-support`; `sentinel-identity` and `identity-collision` under `packages/generate/test`.
- 109 snapshot files promoted, each of one shape: an added `cid-` token per binding, a reference selector collapsing to `.cid-…`, or an empty marker becoming `CSS.make("cid-…", [])`.

Review from `Hash_class.identity_class`, then `Css_file.push`, then `Index` in `packages/generate/generate.ml`. The specificity drop for references is the one behaviour change to weigh; `documents/css-extraction.md` "Identity classes" documents it.

Risk: two-way door for styled-ppx itself; a revert restores chain resolution. For consumers, compound selectors written against a multi-atom reference change specificity, so a monorepo rebuild should be checked with the winner-diff script before the pin bump.


davesnx:

